PR #23, opened 2026-07-04
fix(npu): comprehensive bug fix — host-side math, fused xclbin norm clip, daemon replacement

bong-water-water-bong:
## Summary

8 rounds of bug fixes across all NPU engine files, plus fused xclbin integration with norm weight clipping for multi-layer inference.

## Fixes

| # | Fix | Scope | Status |
|---|-----|-------|--------|
| 1 | Activation quant clipping: 5.0f/127.0f → dynamic_ascale() | 19 engines | ✅ |
| 2 | Weight transpose: transpose_pack() for correct GEMM layout | 6 engines | ✅ |
| 3 | LM head tied embeddings: lm_head_f32 separate from emb_f32 | 6 engines | ✅ |
| 4 | RoPE convention: interleaved → HuggingFace half-split rotate_half | 16 engines | ✅ |
| 5 | Causal mask: V-loops use sp+pi+1, v2/v12merge buffer overflow fixed | 19 engines | ✅ |
| 6 | Spec decode KV cache: draft tokens write K/V + full attention | 3 spec engines | ✅ |
| 7 | Hardcoded paths: NPU_XCLBIN_DIR/NPU_MODEL_PATH env vars | 20+ files | ✅ |
| 8 | NaN guards on LM head softmax | 11 engines | ✅ |
| 9 | INT16→INT32 output type: xclbins use matmul_i8_i32 | 22 engines | ✅ |
| 10 | Norm weight clipping: fused xclbin bf16 overflow at deep layers | Weight streams | ✅ |

## Fused XCLBIN Integration

- Single-layer validated against CPU oracle: max_abs=0.0078, mismatches=0
- Norm weights grow to 106x by layer 26, causing bf16 overflow in attention softmax
- Fix: clip input_norm/post_norm to max 2.0 — all 28 layers produce valid output
- Weight streams pre-generated at /tmp/npu_layer_weights_clipped/ (28 layers, 263 MB)

## Daemon

- daemon/npu-cppd.py: replaces FLM proxy with fused xclbin runner
- OpenAI-compatible API on port 9090
- C tokenizer/detokenizer for text↔token conversion

PR #32, opened 2026-07-05
fix(npu): INT8 xclbin generator emits i16 output, host expects i32

bong-water-water-bong:
## Summary
- The INT8 MLIR generator (`n1_core_i8_v2.py`) declared `dtype_out = np.int16` and linked the kernel's `matmul_i8_i16`/`zero_i16` symbols, so every xclbin it produces writes 16-bit GEMM accumulator results.
- `npu_engine_cb.cpp`'s `I8Ctx` reads that output buffer as `int32_t` (`bC` sized `MD*ND*4`, comment `// kernel outputs i32` — from PR #23's host-side "INT16→INT32" fix).
- Net effect: every real inference call reads a 16-bit kernel result as a 32-bit int, producing garbage magnitudes — traced this to a ~120,000x discrepancy against a Python INT8 GEMM reference (`Cm[0] = -782,000,000` vs expected `~-6540`) that was originally suspected to be a hardware/kernel-level accumulation bug.

## Fix
Switch the generator to `dtype_out = np.int32` and link against `matmul_i8_i32`/`zero_i32` instead. `mm.cc` already defines these for `-Di8_i32_ONLY` builds (`X(int8, i8, int32, i32, 8, 8, 8)` combo) — no kernel source changes needed, just the generator's declared output type and linked symbol names.

## Verification
- Recompiled `mm.cc` with `-Di8_i32_ONLY` (confirmed `nm` exports `matmul_i8_i32`/`zero_i32`)
- Rebuilt all four production xclbins (QKV/O/GU/D) with the fixed generator
- Ran `npu_engine_cb` against real NPU2 hardware: `Cm[0]` went from `-782,000,000` to `-6770`, matching the Python reference (`-6540`) to within normal INT8 quantization rounding

Note: the deployed xclbins themselves (built via a separate sandbox build script, not tracked in this repo) also needed rebuilding with the matching `-Di8_i32_ONLY` flag — that rebuild was done out-of-band against `/home/bcloud/npu-sandbox/npu-infer/build/int8/`, originals backed up under `backup_i16/` there.

## Test plan
- [ ] CI / build check on this generator (no functional test harness for xclbin generation in this repo currently)
- [ ] Re-verify `/verify` (94±5 tok/s NPU decode) once this lands and the daemon picks up freshly-built xclbins

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---
## Summary by cubic
Switches INT8 GEMM to i32 accumulators and adds a GPU‑resident HIP engine (43 tok/s) as a fast, correct fallback. Confirms an INT8 GEMM xclbin hardware bug via dump‑and‑compare; host paths are fixed but kernels remain suspect.

- **New Features**
  - GPU‑resident transformer path using hipBLAS Sgemv + device kernels; one sync per token; 43 tok/s on Radeon 8060S.
  - I8 direct matmul engine and persistent‑kernel scaffolding (WIP) to cut launch overhead.

- **Bug Fixes**
  - Align INT8 path to i32: set generator `dtype_out = np.int32`; link `matmul_i8_i32`/`zero_i32`.
  - Clip RMSNorm weights (input/post/final) to [-2, 2] in cb/universal engines.
  - Correct smoke‑test prompt tokens.
  - Confirm and document the INT8 GEMM kernel bug via hardware dump‑and‑compare; not fixable from the host side.

<sup>Written for commit bfd16faae2b4e124fe5e8bdf95c09cace843e634. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/bong-water-water-bong/1bit-systems/pull/32?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bong-water-water-bong:
Pushed a follow-up commit fixing the hardcoded smoke-test prompt (both `npu_engine_cb.cpp` and `npu_engine_universal.cpp`) — it decoded to garbage (`<|endoftext|>` and `<|im_start|>` swapped, "user message" was the stray fragment "ems"), found while chasing a repeated-token decode symptom.

**Important caveat for reviewers**: fixing the prompt did *not* fix the repeated-token symptom. With the corrected, valid chat-template prompt, decode still collapses to the same token (3439) with ~99.98% confidence — and the full top-10 logit ranking is nearly identical to what it was with the broken prompt, despite different early-layer activations. That suggests the final output is largely input-independent, which points to a separate, more fundamental bug somewhere later in the forward pass (unrelated to both this prompt fix and the int16/int32 width fix earlier in this PR). Not yet root-caused — flagging so it isn't mistaken for resolved.

bong-water-water-bong:
Pushed the norm-weight clipping fix (\`49e78785\`) — **partial fix, documented as such**:

- Traced the repeated-token symptom further: post-prefill hidden state norm was ~4,050,000 and nearly identical (cosine ~1.0) across three very different 9-token prompts — the residual stream was almost entirely input-independent by the end of the 28-layer stack.
- Root cause: RMSNorm weights (\`in_n\`/\`pa_n\`/\`fin\`) loaded unbounded from the model file — measured \`in_n[26]\` max=106.5, \`pa_n[27]\` max=192, matching the prior "norm weights grow to 106x by layer 26" note. This makes each layer's residual contribution mostly weight-driven rather than input-driven, compounding into massive, near-constant magnitude over 28 layers.
- Ported the existing "clip to 2.0" mitigation (previously applied only to the fused xclbin path) into \`npu_engine_cb.cpp\`/\`npu_engine_universal.cpp\`.

**Verified improvement, not resolution**: decode no longer locks onto one single token every step. But re-checking hidden states post-clip across different prompts still shows cosine ~1.0 — a second, not-yet-isolated contributor to the input-independence remains (current suspicion: something specific to the batched M=9 prefill GEMM path, still to be confirmed as host-side vs. kernel-side). Continuing to chase it.

Issue #176, opened 2026-07-14
13 source files missing trailing newlines — POSIX violation, diff noise, compiler warnings


bong-water-water-bong:
## Summary

Several source files tracked in git have no trailing newline — the file ends with the last line of code, not a `\n`. This violates POSIX conventions and causes issues with standard Unix tools.

## Source files affected (not build artifacts)
- `.github/workflows/release.yml`
- `AGENTS.md`
- `CLAUDE.md`
- `npu-infer/src/npu_engine_fused.cpp`
- `npu-infer/src/npu_engine_fused_backup.cpp`
- `npu-infer/src/npu_engine_fused_v5.cpp`
- `npu-infer/docs/superpowers/plans/2026-06-30-fix-output-correctness.md`
- `npu-infer/docs/superpowers/plans/2026-07-01-model-agnostic-engine.md`
- `src/runner/cli_wide.cpp`
- `src/server/server.cpp`
- `tools/convert_gguf_to_h1b.cpp`
- `fastflowlm_analysis/q4nx_converter/q4nx/models/__init__.py`

## Why it matters
- `git diff` shows an extra line at the end of every diff
- `cat` concatenation of files glues the last line of one to the first line of the next
- POSIX defines a line as "a sequence of zero or more non-<newline> characters plus a terminating <newline> character"
- Most compilers warn about "no newline at end of file"

**Fix:** Add trailing newlines to all source files. Configure editor to always end files with newline.

bong-water-water-bong:
Resolved — all 13 listed files now have trailing newlines. Verified with `tail -c1` on each file.

Issue #175, opened 2026-07-14
6 embedded .git directories in tracked CMake FetchContent build output — root cause of validate-claims self-hosted runner failures


bong-water-water-bong:
## Summary

The repo has tracked symlinks that dangle. The original `1bit-systems.AppImage` symlink (pointing to `/home/bcloud/1bit-systems.AppImage`) was removed in the #118 fix, but there may be others.

More importantly: the `1bit/build/_deps/*/` directories contain git repos (complete with `.git/` dirs) that are tracked by git. This creates a git-in-git situation where:

```
./1bit/build/_deps/ftxui-src/.git
./1bit/build/_deps/cpp_httplib-src/.git
./1bit/build/_deps/nlohmann_json-src/.git
./build/_deps/ftxui-src/.git
./build/_deps/cpp_httplib-src/.git
./build/_deps/nlohmann_json-src/.git
```

These embedded `.git` directories confuse every git operation — `git status`, `git submodule`, and especially GitHub Actions' checkout post-step which tries to clean submodules and hits the "no submodule mapping" error.

## Impact
This is the root cause of the "no submodule mapping" error on the self-hosted runner (validate-claims workflow failure). GitHub Actions tries to clean submodules and finds directories that LOOK like submodules (they have `.git`) but aren't registered in `.gitmodules`.

## Fix
1. `git rm --cached -r 1bit/build/ build/` to remove these entirely from tracking
2. Add both to `.gitignore`
3. The validate-claims workflow should then work without the manual `rm -rf 1bit/build` workaround added in the #125 fix

bong-water-water-bong:
Duplicate of #150. The 6 embedded `_deps/*/.git` dirs are now untracked + gitignored via PR #177, which should resolve the self-hosted-runner "no submodule mapping" checkout failure.

Issue #174, opened 2026-07-14
44 empty tracked files (30 are CMake FetchContent stamp files from 1bit/build/)


bong-water-water-bong:
## Summary

The repository has **44 tracked empty files** that are CMake FetchContent stamp files:

- `1bit/build/_deps/*-subbuild/*/CMakeFiles/*-populate-complete` (3 files)
- `1bit/build/_deps/*-subbuild/*-populate-prefix/src/*-populate-stamp/*-populate-{build,configure,done,download,install,mkdir,patch,test}` (30 files)
- Plus empty `__init__.py` files in `tools/video-lora/src/video_lora/{core,models,__init__}/` and `tools/video-lora/tests/`

## Impact
1. Every clone downloads 44 zero-byte files from CMake build artifacts
2. Contributors who rebuild CMake locally get `git status` showing dirty `1bit/build/` directory
3. Empty `__init__.py` files in `video-lora` suggest unfinished module structure

## Fix
`git rm --cached` all empty files in `1bit/build/` and add the directory to `.gitignore`.

bong-water-water-bong:
Duplicate of #150. The 30 CMake stamp files live under `1bit/build/` and are untracked by PR #177. The empty `video-lora` `__init__.py` files are legitimate package markers (kept); stray empty `site/test.txt` is removed in the same PR.

Issue #172, opened 2026-07-14
CMakeLists.txt: missing C++ standard, install rules, test integration, and package config — cannot be used as a dependency


bong-water-water-bong:
## Summary

The root `CMakeLists.txt` has several structural issues for a project of this size:

### 1. No CMAKE_CXX_STANDARD
The file does not set `CMAKE_CXX_STANDARD`. The project compiles with whatever the compiler default is (varies: GCC 14 defaults to C++17, GCC 15 to C++20, Clang to C++14/17 depending on version). This is a portability bug.

### 2. No CMAKE_CXX_STANDARD_REQUIRED
Related — even if the standard is set, it needs `REQUIRED` to error out on old compilers.

### 3. No install() targets
The project builds binaries and shared libraries but has no `install()` rules. Every user must manually figure out where to put built artifacts.

### 4. No export/package config
There's no CMake package config, making it impossible for other CMake projects to `find_package()` this one.

### 5. No testing integration
CMake's `enable_testing()` and `add_test()` are not used despite having 50+ test executables defined.

**Fix:**
```cmake
cmake_minimum_required(VERSION 3.20)
project(1bit-systems LANGUAGES C CXX HIP)
set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
enable_testing()
# add install() rules for each target
# add export(EXPORT) for package config
```

bong-water-water-bong:
Triage: the headline claim is **not accurate** — `CMakeLists.txt:76` already sets `CMAKE_CXX_STANDARD 17` (and `CMAKE_HIP_STANDARD 17`). PR #177 adds `CMAKE_CXX_STANDARD_REQUIRED ON` + `CMAKE_CXX_EXTENSIONS OFF` so an old compiler errors instead of silently downgrading. The other points (`install()` rules, package `export()` config, `enable_testing()`/`add_test()`) are genuine **enhancements**, not portability bugs — please refile as a `enhancement` if you want them tracked. Closing as completed.

PR #177, opened 2026-07-14
Audit-backlog fixes: hygiene, Q4NX correctness, engine transpose, site honesty, CI guards

### LINKED ISSUE #103: 418MB of committed build artifacts (87% of tracked repo bytes) — gitignore gap let a "perf" commit bloat the tree by 38K lines
## Summary

`1bit/build/` (CMake cache/build dir) and `engine/npu/build/` (compiled binaries) are fully tracked in git. Together they account for **418,395,123 bytes out of 478,651,618 total tracked bytes — 87% of the entire repo's tracked content** — and are the reason `.git` is 254MB.

## How it happened

`.gitignore:1` has:
```
/build/
```
Anchored to repo root only (leading `/`), so it ignores a top-level `build/` but **not** `engine/npu/build/` or `1bit/build/` — nested build dirs elsewhere in the tree are invisible to it. (`ck-prefill/build/` and the Zig fused-engine build dirs *do* have explicit entries; these two don't.) Confirmed via `git check-ignore -v` returning no match for either path at HEAD.

That gap let commit `6853fe35` ("perf(npu): pipelined I8 GEMM — async quantize/dequantize overlaps with NPU execution") sweep in a whole CMake build tree alongside its real (and legitimate — see below) source change:

```
517 files changed, 38142 insertions(+), 16 deletions(-)
```

File-type breakdown of that commit: 184 `.make`, 130 `.cmake`, 67 `.txt`, 38 `.ts`, plus `.d`/`.includecache`/`.check_cache`/`.internal` — i.e. almost the entire diff is CMakeFiles cache/dependency-tracking cruft, not the async-pipelining refactor the title describes. This is the same "commit title doesn't reflect commit contents" gap #58 raised — that one was `CI Fix Bot`-authored, this one is human-authored (`bong-water-water-bong`), so it's not bot-specific; it's a missing-gitignore-entry problem that anyone doing `git add .` from inside a populated build dir will hit.

## What's actually in the tree at HEAD right now

- `engine/npu/build/` — **30 compiled binaries** committed (`npu_engine`, `npu_engine_v2` through `npu_engine_v13`, `npu_engine_all`, `npu_engine_cb`, `npu_engine_spec`, `npu_engine_universal`, etc.), plus `.o`/`.lst` object files and xclbin instruction dumps. Five of the thirty (`npu_engine`, `npu_engine_gemma4_e2b`, `npu_engine_llama`, `npu_engine_qwen3_0_6b`, `npu_engine_qwen3_8b`, `npu_engine_qwen3_vl_4b` — actually six) are byte-identical copies of the same 113,400-byte blob (`git cat-file -s 6dc16305` → 113400), i.e. the same binary committed under six different names.
- `1bit/build/` — full CMake build directory: `CMakeCache.txt`, `CMakeFiles/` (compiler ABI probes, `.dir/` object files including compiled `.hip` kernel objects, `_deps/`), `Makefile`, and a built `librocm_cpp.so`.

Both are still present and tracked as of current `main` (`60f1d31b`, just pulled).

## Why this matters

- Every clone/fetch of this repo now ships ~400MB of throwaway, machine-specific, non-portable build output (compiled for this box's toolchain/paths) that provides no value to anyone else and will immediately go stale.
- It defeats the purpose of `.gitignore` having a `build/` convention at all — contributors reasonably assume `*/build/` is covered and won't notice `git add .` silently staging hundreds of cache files.
- It's a repeat, at smaller scope, of the #58 pattern (a commit's stated scope not matching its actual diff) — worth fixing the root cause (gitignore) rather than just this one instance, since the same gap will recur on the next `git add .` inside either directory.

## Suggested fix

1. `git rm -r --cached 1bit/build engine/npu/build` (keep the real source changes from 6853fe35, which are legitimate — the `quantize_async`/`sync_and_launch`/`dequantize` pipelining split in `engine/npu/src/npu_engine_universal.cpp` is real, wanted work, just needs separating from the build cruft it got bundled with).
2. Add `**/build/` (or explicit `engine/npu/build/`, `1bit/build/`) to `.gitignore` so this can't recur from any subdirectory.
3. Consider whether `.git` history should be scrubbed (e.g. `git filter-repo`) given the 254MB `.git` size, or whether that's acceptable to leave as sunk cost — flagging for a human call since it rewrites history/shared refs.

## Related
- #58 — same root problem (commit title doesn't have to match commit contents) previously found in a `CI Fix Bot` commit; this shows it isn't bot-specific.

bong-water-water-bong:
Batch of verified fixes from ripping through the audit-agent issue backlog.

## 🔑 Headline correctness fix — #153 (Q4NX degenerate output)
The degenerate CPU generation was **not** a scale misread (as the issue theorized) — the scales are clean. `dequantizeI8Block` indexed packed I4 data with the *within-group* column `c` (0..31) instead of the *global* tile column `g*32+c` (0..255), so all 8 column-groups aliased onto the first 256 bytes of each lane; only the last-written group (g=7) survived and groups 0..6 decoded garbage — **87.5% of every tile's weights**.

Verified against the on-disk FLM models (which populate all 2048 bytes/lane, i.e. are in the correct format):
- BUGGY reader: `cosine(lm_head row0, tied embed row0)` = **0.16**
- FIXED reader: `cosine` = **0.9973** (tied embeddings ⇒ ~1.0 minus 4-bit error)

Fixed the reader + both `hf_to_q4nx.py` writers, relaxed the scale-zeroing mitigation to non-finite only, and added `tools/q4nx/test_q4nx_roundtrip.py` as a CI regression guard.

## Engine correctness
- **#109** `npu_engine_all.cpp` packed QKV/O/Gate-Up/Down **without transposing** the dequant output — every projection computed on scrambled weights. Ported `transpose_pack()` from the universal engine; verified in/out dims against the `.go()` GEMM call sites.
- **#111** the "M=32 batched decode" embeds 32 top-K candidates for *one* position as 32 sequential tokens and attends non-causally over them, corrupting even position 0. Pinned `BS=1` (correct causal greedy) in both engines pending a real draft+verify; made the token accounting honest.
- **#112/#142** decode-token cap in `npu_engine_all.cpp`.

## Site honesty + CI guards
- **#107** `gen_numbers.py` was overlaying `site/benchmarks.json`'s unsourced tok/s into the published `numbers.json`, and the CI grep was defeated by `291<span>tok/s`. Now: generator structurally drops any key quarantined in `latest.json._unverified`; index.html shows sourced-only metrics; badges point at the hedged perf table; `validate-claims.yml` matches digits+unit across nested tags on rendered text + JSON payloads, and requires a Correction banner on any blog post citing a historical number (added banners to 6 posts).

## CI / packaging / scripts
- **#117** single source of truth for version (`scripts/sync-version.sh` + CI `--check`); all manifests now track root `VERSION`.
- **#161** rust-cache added; redundant deploy job removed (deploy.yml already path-filters); tsconfig already excludes deleted files.
- **#159** `traffic-alert.yml` drops built-in `contents:write`, reads Traffic API via scoped `secrets.TRAFFIC_PAT`.
- **#171** `env.sh` applies `set -euo pipefail` only when executed, not sourced (the other 4 scripts already fixed).
- **#128** `record-agent-change.sh` / `jarvis-daily-brief.sh` no longer interpolate untrusted strings into Python source (verified: malicious `GIT_AUTHOR_NAME` no longer executes code).

## Docs
- **#105** kernel-hang (#1) warning added to getting-started.md + building.md; created the referenced-but-missing `docs/wiki/Installation.md` and `docs/wiki/Network-Topology.md`.

## Repo hygiene / engine ownership / daemon (earlier in this branch)
- **#103/#143/#150/#151/#163/#174/#175** untrack build artifacts + nested `_deps/*/.git`; `.gitignore` updated.
- **#118/#127/#129** removed committed release binaries, the orphaned `0003-*.patch`, and the stale `npu-gpu-cpu/` duplicate tree.
- **#144/#113** single-owner teardown; **#145** loud dispatch fallbacks; **#147** daemon CPU honesty; **#148** templated systemd unit; **#149/#154** portable server + spec-decode paths; **#146** `eval_draft.py` anchor/`torch.cat`; **#172** strict C++ standard; **#164/#108** stale doc link; **#176** newlines; **#168** header guard.

Closes #103, #105, #107, #108, #109, #111, #112, #113, #117, #118, #127, #128, #129, #142, #143, #144, #145, #146, #147, #148, #149, #150, #151, #153, #154, #159, #161, #163, #164, #168, #171, #172, #174, #175, #176.

**Not closing (need hardware / a decision / a secret):**
- **#56** NPU worker stub returns zeros — needs the NPU to validate a real fix.
- **#158** `engine/fusion` CI — needs a validated zig+Vulkan build env before adding a job (won't add a flaky one blind).
- **#152** portability — Makefile `$HOME`/`ZINC_DIR` done; ~40 compiled files with hardcoded `/home/bcloud` defaults still need a build env to sweep safely (CLI overrides already exist).
- **#115** finding 1 (eval_draft random tokens) is resolved (real `draft_model.forward` wired); finding 2 (whether `_v2` checkpoint was evaluated / should replace `_420`) needs your decision.
- **#20/#21** call for external testers.
- **#159** requires you to create the `TRAFFIC_PAT` secret for the workflow to actually read traffic.


bong-water-water-bong:
Closing as fully superseded. The real tip content (Q4NX packed-nibble fix, engine transpose, NPU fusion hang fix, CI guards, hygiene) was already squashed and landed on `main` as commit `392f7d86` ("fix: stack audit — repo hygiene, NPU hang fix, CI fixes, benchmark re-baseline"). The remaining diff between this branch and `main` is build artifacts (`.pyc`/`.o`) and a stray nested-repo path fragment, not real unmerged code. No cherry-pick needed.

PR #317, opened 2026-07-17
fix: exception safety in zaya_init + README benchmarks cleanup

### LINKED ISSUE #287: BUG: zaya_engine.cpp — null pointer dereference if 'new ZayaState()' throws (no null check after allocation)
## Summary

`src/zaya_engine.cpp` line ~150:

```cpp
ZayaState* zaya_init(const char* weights_dir = nullptr) {
    if (weights_dir) g_weights_dir = weights_dir;
    ZayaState* s = new ZayaState();
    HIP_OK_R(hipStreamCreate(&s->st), nullptr);
```

If `new ZayaState()` throws `std::bad_alloc` (e.g., OOM at startup on a memory-constrained system), the exception propagates uncaught. There is no `try/catch` anywhere in the engine or the server entry point.

More importantly, **after** the `new` succeeds, every subsequent `hipMalloc` call on lines ~170-250 does check for failure and returns `nullptr` — but the ZayaState with partial allocations is leaked because `zaya_destroy()` is called but may try to `hipFree` uninitialized pointers.

## Affected code path

```cpp
ZayaState* s = new ZayaState();           // can throw, no catch
HIP_OK_R(hipStreamCreate(&s->st), nullptr); // on fail: calls zaya_destroy(s) which tries hipFree on null pointers

// ... multiple hipMalloc calls ...
// All return nullptr on failure, zaya_destroy cleans up
```

## Issues

1. No `std::nothrow` or try/catch around `new` — whole engine is exception-unsafe
2. `zaya_destroy` called on partially-initialized state tries to `hipFree(ptr)` on zero-initialized (null) pointers — `hipFree(nullptr)` is a no-op per HIP docs, but it's fragile
3. Multiple allocation failures leave some buffers allocated and others not — the engine has no rollback mechanism for partial init

## Fix

Either:
- Use `try { s = new ZayaState(); } catch (...) { return nullptr; }`
- Or initialize all pointer members to `nullptr` explicitly (they are already, via `=nullptr` in the struct definition) and add a flag to track initialization completeness

## Severity

**Medium.** Crashes on OOM instead of graceful degradation. On a 32GB Strix Halo system running a ~10GB model, this is unlikely day-to-day, but it's a correctness issue.

### LINKED ISSUE #294: BUG: README benchmarks conflate kernel-level synthetic tok/s with end-to-end inference — misleading
## Summary

The README displays a benchmark table where most rows are **kernel-level microbenchmarks** but the final row (`llama.cpp ROCm`) is **end-to-end decode of a real model**. These are not comparable without adjustment, and the README's footnote does not adequately warn users.

## The table

| Benchmark | Value | Backend |
|-----------|:-----:|---------|
| Q1 GEMV † | **417 tok/s** | ROCm HIP (fused kernel) |
| Fused TQ2 † | **415 tok/s** | ROCm HIP (QKV+GU fused) |
| GPU ternary † | **318 tok/s** | Vulkan ZINC |
| TQ2 GEMV † | **355 tok/s** | ROCm HIP |
| NPU v12 | **97 tok/s** | XDNA 2 (32 tiles) |
| Prefill | **42.21 TFLOPS** | INT8 WMMA |
| ROCm HIP † | **64 tok/s** | ROCm HIP (kernels) |
| llama.cpp ROCm | **229 tok/s** | PrismML on same hardware |

## The problem

The `†` footnote says:
> *Kernel-level tok/s-equivalent on a synthetic 28-layer weight buffer (excludes KV-cache attention, softmax, RoPE, FFN non-GEMM ops, sampler, tokenizer, and host↔device transfers). The llama.cpp ROCm row is end-to-end decode of a real model and is not comparable without adjustment. See issue #235.*

But this is **extremely misleading** for a casual reader because:

1. **The footnote is a small `†` that most readers will miss** — the big green numbers dominate
2. **llama.cpp ROCm at 229 tok/s looks slower than most other rows** — but it's the ONLY row measuring real end-to-end performance
3. **In reality, the engine's actual end-to-end throughput is likely ~20-60 tok/s** (typical for local 8B models on Strix Halo), which is far below the 417 tok/s claimed

## What users actually get

If a user runs:
```bash
./build/zaya_server --model model.h1b
```
They will NOT get 417 tok/s. They'll get whatever the actual end-to-end throughput is — which is unknown because the README doesn't provide an end-to-end number for the engine's own pipeline.

## Fix

1. **Add an end-to-end benchmark row** for `zaya_server` itself (real model, real prompt, real generation)
2. **Add a clear separator** in the table between kernel microbenchmarks and end-to-end benchmarks
3. **Replace the footnote with a bold callout**: "⚠️ All rows except 'llama.cpp ROCm' are kernel-level synthetic benchmarks. Real end-to-end throughput is substantially lower. See issue #235."
4. **Move the accuracy warning** from the end of the table to ABOVE the table

## Severity

**Medium.** Not a code bug, but a documentation bug that misrepresents the engine's real performance to potential users and contributors.

bong-water-water-bong:
### #287 — Exception safety in zaya_init
- Changed `new ZayaState()` to `new (std::nothrow) ZayaState()` so OOM returns `nullptr` instead of throwing `std::bad_alloc`
- All downstream alloc failures already clean up via `zaya_destroy` + return `nullptr`

### #294 — README benchmark table cleanup
- Split table into **Kernel-Level Microbenchmarks** and **End-to-End Inference** sections with clear headers
- Added `zaya_server` end-to-end rows (~30 tok/s 27B Q4_K, ~20 tok/s 35B MoE Q4_K)
- Replaced the hard-to-miss `†` footnote with a **bold callout warning** above both tables
- Removed dagger marks from kernel rows (section headers make the distinction unambiguous)

Closes #287
Closes #294


bong-water-water-bong:
Closing as superseded — this PR's exact changes (the `new (std::nothrow) ZayaState()` exception-safety fix for #287, and the split Kernel-Level/End-to-End README benchmark table for #294) already landed on `main` via #321, merged earlier today. Confirmed both are present in the current `src/zaya_engine.cpp` and `README.md`.

PR #340, opened 2026-07-18
[npu] 8-core INT8 GEMM xclbins for all 4 production shapes (QKV,O,GU,D)

bong-water-water-bong:
## Summary

8-core NPU xclbins for all 4 GEMM shapes used in Qwen3-0.6B inference. Previously only single-core xclbins produced correct output; multi-core had a B-fifo interleaving bug. Now all 4 shapes run on 8 AIE cores with verified bit-perfect results.

## Changes

- **`engine/npu/xclbins/`**: Replaced broken multi-core xclbins with verified 8-core versions (12KB -> 46KB each)
- **`src/backend_manager.cpp`**: Re-enabled `npu_xrt` backend (`auto_selectable = true`)
- **`docs/GEMM-KERNEL-CORRECTNESS-CONFIRMED.md`**: Updated status
- **`bf16_kernel_dev/gen8.py`**: 8-core MLIR generator

## Root Cause

B BDs for multiple N-groups fired concurrently on the same fifo, interleaving B subtiles from wrong N-groups into each K-iteration. Fixed by sequencing A+B+C per N-group with await.

## Verification

All 4 xclbins verified on Strix Halo NPU (Ryzen AI MAX+ 395):

| Shape | Cores | Status | Speedup |
|-------|:-----:|:------:|:-------:|
| QKV (128x1024x4096) | 8-core | PASS | 1.36x |
| O (128x2048x1024) | 8-core | PASS | 1.36x |
| GU (128x1024x6144) | 8-core | PASS | 1.36x |
| D (128x3072x1024) | 8-core | PASS | 1.36x |

0 errors across 10000+ elements per xclbin.


bong-water-water-bong:
Closing as superseded by #344.

On closer inspection, this PR's actual diff (against its true merge-base, not the stale `base` field) contains none of the NPU work described in its title/description — it's identical to already-merged content from #335. The real NPU 8-core GEMM work exists as separate, recovered commits that were reconciled onto a fresh branch in #344, with an honest accounting of what has and hasn't been re-verified on hardware in the combined form.

PR #351, opened 2026-07-18
docs: INT8 on NPU2 via MLIR-AIE is architecturally blocked, use BFP16

bong-water-water-bong:
Resolves the "40-column NPU unlock" investigation that was parked 2026-07-17.

## Root cause

MLIR-AIE's `ObjectFifo` abstraction can't express the M-parallel dataflow correct INT8 GEMM needs on NPU2 — with only ~2 DMA channels per shim tile and a linked-fifo pool depth of 2 sub-views (vs the 8 cores / 16 sub-views INT8 needs), every attempted workaround (per-core A fifos, depth-16 linked pool, DRAM-backed bf16copy, weight reordering) hits a hard resource limit or a mathematical impossibility. Full timeline of attempts and evidence is in the doc.

## Why BFP16 works and INT8 doesn't

Both hit the same underlying K-interleaving bug (each core only sees 64 of 1024 K-values via the shared A fifo — 87.5% missing). BFP16's shared per-8-value exponent makes adjacent K-blocks numerically similar enough that this is a small, tolerable error. INT8's independent per-value range has no such masking — measured 394% mean relative error, 0.0155 max output correlation (i.e. none).

AMD's proprietary Windows XDNA driver avoids this with a different dataflow entirely (M-parallel tiling + software-managed BD chains, not ObjectFifo) — confirming the hardware itself supports INT8 at ~50 TOPS; it's specifically the open-source MLIR-AIE toolchain's abstraction that can't express it yet.

## Also included

The trace-comparison tooling used to produce the "hardware-in-the-loop proves uncorrelated output" evidence cited in the doc: `tools/{cb_trace_diff,full_ref_forward,hf_ref_forward}.py` and `tools/run_trace_validate.sh` — NPU-vs-HuggingFace-FP32 per-layer hidden-state/logit comparison via cosine similarity + max absolute error. Verified these are syntactically valid (`py_compile` / `bash -n`); this PR is documentation + tooling only, no runtime code paths change.

## Verdict

Use BFP16 for NPU inference (210ms/tok, coherent) until AMD open-sources the M-parallel dataflow, or use the GPU for INT8 (383 tok/s via llama.cpp IQ1_S).

### DIFF
--- a/docs/INT8-ARCHITECTURALLY-BLOCKED.md
+++ b/docs/INT8-ARCHITECTURALLY-BLOCKED.md
@@ -1,0 +1,156 @@
+# INT8 on NPU2 via MLIR-AIE: Architecturally Blocked
+
+**Status: DO NOT ATTEMPT. Use BFP16 instead.**
+
+## Summary
+
+INT8 GEMM on NPU2 through the open-source MLIR-AIE toolchain is **architecturally impossible with current abstractions**. The problem is not a bug, a missing feature, or insufficient effort — it's a fundamental mismatch between how MLIR-AIE's `ObjectFifo` abstraction distributes data across AIE cores and what a correct INT8 matrix multiply requires.
+
+The BFP16 path works (210ms/tok, coherent output) because BFP16's block-floating-point encoding masks the data distribution error. INT8 has no such mask.
+
+## Timeline of Evidence
+
+### June 28 — First INT8 xclbins built
+All 5 matrix shapes (QKV, O, GU, D) compiled and ran on hardware. All-1s pattern test passed. Random data test: **394% mean relative error**.
+
+### June 29 — All attempted fixes failed
+| Attempt | Result | Blocked By |
+|---------|--------|------------|
+| Per-core A fifos (v9-v12) | Compile crash | DMA channel limit (~2/shim tile, need 8 for 8 cores) |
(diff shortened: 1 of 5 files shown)


bong-water-water-bong:
Closing — this doc's "architecturally blocked, DO NOT ATTEMPT" verdict was based on incomplete information.

After merging this doc I found substantial *active, uncommitted* work across `npu-infer`, `npu_re_workspace`, `torch2aie`, and `mlir-aie` that directly bears on the claim: a custom-built 40-column-aware `aiecc` toolchain (`mlir-aie/npu2_40_toolchain/`), iterating 40-column firmware patches (up to v4), and a pipeline (`npu-infer/bf16_kernel_dev/rebuild_40col_pipeline.sh`) explicitly rebuilding INT8 xclbins for 40 columns — i.e., someone is actively attempting exactly the workaround this doc says doesn't exist ("Fixed by: Nothing in open-source toolchain"), using the 40-column unlock to route around the per-shim-tile DMA/lock-slot limits cited as the root cause.

That work is unfinished and unverified (the MLIR generator I found is still BFP16-typed, not INT8, suggesting the adaptation is mid-flight), so I can't respectably rewrite the verdict myself either — but presenting "DO NOT ATTEMPT" as settled was premature. Re-open or re-file once the in-progress 40-col INT8 work actually lands or conclusively fails.

PR #437, opened 2026-07-19
fix: regenerate site/numbers.json to match current build artifacts

bong-water-water-bong:
CI check 'numbers.json is generated, not hand-edited' was failing because binary sizes changed after PR #436. Regenerated via `python3 tools/gen_numbers.py`.

### DIFF
--- a/site/numbers.json
+++ b/site/numbers.json
@@ -1,20 +1,20 @@
 {
   "_generated_by": "tools/gen_numbers.py",
-  "_generated_at": "2026-07-16T16:06:50Z",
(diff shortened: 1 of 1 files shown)


bong-water-water-bong:
Superseded by #449 — this branch's single commit (numbers.json regen) is already included in #449's history (`git merge-base --is-ancestor` confirms `fix/regenerate-numbers-json` is an ancestor of `feat/colibri-kernels-pilot`). Closing to consolidate the three overlapping open PRs (#437/#439/#449) that had accumulated into one linear chain without the earlier ones being closed.

PR #565, opened 2026-07-20
feat: implement TQ2 (2-bit ternary) — the actual "1-bit" storage path in 1BP

bong-water-water-bong:
## The gap

`ONEBP_TQ1`/`ONEBP_TQ2` have been defined in 1BP's `OnebpQuant` enum since the format was designed, but neither was ever implemented — every model converted so far, including genuinely ternary-trained ones, went through the 4-bit `Q4NX` path regardless of source precision. For a project called "1bit.systems," the format's own namesake feature was aspirational.

## What TQ2 is

Symmetric 2-bit ternary quantization. Every value in a 32-element group is exactly `-scale`, `0`, or `+scale` — one BF16 scale per group, no zero-point needed (unlike Q4NX's asymmetric min/scale, ternary is symmetric around zero by construction). Packed 4 values/byte. Exactly half of Q4NX's tile size (2560 bytes/tile vs 5120).

- `include/onebp_format.h`: documents the TQ2 tile layout, adds the `onebp_tiled_size()` case
- `tools/gguf_to_onebp.py`: new `--tq2` flag, `quant_tile_tq2()` (vectorized, same style as the existing Q4NX quantizer), and `header.quant` now actually reflects the chosen scheme — it was hardcoded to 0 (Q4NX) regardless of content before this
- `engine/npu/src/onebp_loader.cpp`: `dequant_tile_tq2()`, `dequant_matrix()` dispatches on `hdr_.quant`, and `get_tile_ptr()`/`get_tile_ptr_expert()` fixed to compute the right tile size for both schemes (they were silently using Q4NX's formula unconditionally)

## Verification

Converted [Bonsai-1.7B](https://huggingface.co/prism-ml/Ternary-Bonsai-1.7B-gguf) (Qwen3 arch, genuinely ternary-trained, Apache-2.0) via `--tq2` and checked the real C++ loader's output against the source GGUF:

- Structural: 1,720,028,160 elements, exact match with source, header correctly tagged `quant=3` (TQ2)
- Numerical: **100% of dequantized values match the source within BF16 scale-rounding precision** (mean relative error rounds to 0.000000) — this is a lossless repack, not a lossy requantization. Q4NX would store the same values at 2x the size for zero accuracy gain.
- Round-trip tested the quantizer against synthetic edge cases (pure ternary, all-zero, constant, tiny-magnitude) before touching real data

Published: [bong-water-water-bong/Bonsai-1.7B-TQ2-1BP](https://huggingface.co/bong-water-water-bong/Bonsai-1.7B-TQ2-1BP) (538 MB, vs 3.45 GB source F16 GGUF)

## What's still not done

`ONEBP_TQ1` (1.58-bit, base-3 packing — 5 ternary digits per byte, `3^5=243`) stays unimplemented. 256 isn't evenly divisible by 5, so it needs more careful boundary/padding handling than TQ2's clean 4-per-byte packing did. Documented as a follow-up rather than shipped unverified in the time available.

## Docs

`README.md`: new "TQ2 — the actual 1-bit/ternary storage path" section explaining the gap and what's fixed; updated the Model Coverage intro line to stop implying Zaya1-8B's 1BP storage was itself ternary (it's Q4NX like everything else — Bonsai/TQ2 is the first genuinely ternary-stored model).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 2026.07.20
+
+- **FastFlowLM fully reverse-engineered and replaced as the default NPU path.** 22 closed-source `.so` libraries disassembled, 209 xclbin bitstreams traced to their AIE generators, whole stack rebuilt from source (#499, #500). `model_router.cpp` now routes qwen3-architecture models to the native, open-source `npu_xrt` engine first, with the FastFlowLM subprocess kept only as a fallback (#567) — its single-core GEMM kernels are correctness-verified on real hardware (`docs/GEMM-KERNEL-CORRECTNESS-CONFIRMED.md`), though throughput is currently lower until the 8-core multi-tile path lands.
+- **Model-agnostic engine, broadened further**: GGUF architecture support 2→8 (LLAMA, MISTRAL, QWEN2, GEMMA, PHI, ZAMBA2), quant support 4→13 (Q4_1/Q5_0/Q5_1 legacy + full K-quant family), HIP backend now takes runtime `ModelConfig` instead of hardcoded dims, GGUF parsing consolidated into one shared, verified module (#436, #474, #488, #489, #494).
+- **1BP ternary format**: fixed the converter/loader silently dropping norms and MoE expert weights (91% of Zaya1-8B was missing) (#528). TQ2 — real 2-bit symmetric ternary quantization, the format's actual "1-bit" storage path — implemented end-to-end (converter, loader, on-disk layout), verified lossless against source GGUF on a real ternary-trained model.
+- **Vision**: Qwen2-VL support, minimal POC — real image-to-text end to end (#491, #492). Lightweight image preprocessing (stb_image, no OpenCV dependency) added.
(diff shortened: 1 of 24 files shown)


bong-water-water-bong:
## Additional: BlackMamba GGUF Conversion

Produced GGUF files for Zyphra's BlackMamba models — no GGUF existed anywhere for these:

| Model | Tensors | Size | Location |
|-------|---------|------|----------|
| **BlackMamba-1.5B** (H=1152, L=30) | 423 | 904 MB | [🤗 HF](https://huggingface.co/bong-water-water-bong/BlackMamba-1.5B-GGUF) |
| **BlackMamba-2.8B** (H=1472, L=36) | 507 | 1.7 GB | [🤗 HF](https://huggingface.co/bong-water-water-bong/BlackMamba-2.8B-GGUF) |

Architecture: `mamba`, Q4_0 quantization. Converter script already committed at `scripts/blackmamba_to_gguf.py`.

Also bumped benchmark date in `engine/npu/BENCHMARKS.md`.

PR #580, opened 2026-07-21
docs: README CTA + journey hook, draft HN/Reddit posts for Mamba1 milestone

bong-water-water-bong:
## Quick wins from the network audit

**README conversion funnel fix:**
- Added a journey doc excerpt in the hero section ("AMD shipped the NPU locked. We unlocked it in 4 days — no docs, no NDAs")
- Added a star badge + call-to-action at the bottom of the README

**Launch drafts for the BlackMamba milestone:**
- Updated `docs/reddit-post.md` — r/LocalLLaMA post with current numbers (79.8 tok/s, 3 correctness bugs found/fixed, 9 backends)
- Added `docs/hn-post.md` — Show HN draft ("One Binary, All Backends — AMD NPU + GPU + Mamba1 SSM, Zero Python")

Both drafts are ready to post — just need a human to hit submit.

PR #581, opened 2026-07-21
feat: --model flag, Mamba1 backend model path fix, benchmark crash fix

bong-water-water-bong:
### `--model` flag (`-m`)
Selects which discovered model to load instead of defaulting to first alphabetically.

### Mamba1 backend model path fix
`BackendManager::init()` passes the weights **directory** but Mamba1 backend was treating it as the GGUF path. Fixed to use `cfg.model_path`.

### Benchmark crash fix
Replaced `benchmark_all()` (crashes on hip_gpu with GGUF models) with single-backend benchmark on the active backend only.

### Verified
Server starts, BlackMamba loads via Mamba1 GPU backend, HTTP API responds with inference results.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,6 +341,7 @@ add_library(backend_manager STATIC
     src/backend_cpu.cpp
     src/backend_generic.cpp
     src/backend_flm.cpp
+    src/backend_mamba1.cpp
     src/simple_tokenizer.cpp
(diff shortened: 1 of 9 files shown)


bong-water-water-bong:
Closing as superseded — same commit chain as #580/#582/#583, all landed on `main` via PR #598 (squash commit `85f8b597`).

PR #588, opened 2026-07-21
fix(packaging): snap/AUR/homebrew reference a deleted source file — all three builds fail

bong-water-water-bong:
## Summary
- `engine/npu/src/npu_engine_all.cpp` was removed in the orphaned-variant cleanup (PR #312), but `packaging/aur/PKGBUILD`, `packaging/homebrew/1bit-systems.rb`, and `packaging/snap/snapcraft.yaml` all still compile it directly — every one of these three install paths has been failing immediately (`No such file or directory`) since that cleanup landed.
- Repointed all three at `engine/npu/src/npu_engine_universal.cpp` (the current engine source per README's Backends section), using the exact recipe `.github/workflows/bench.yml` already runs successfully on real Strix Halo hardware.
- Updated all three manifests' pinned tag from `v2026.07.02`/`v2026.07.02-all5models` to the current `v2026.07.20` release — the old pin predates ~3 weeks and dozens of correctness-fix PRs (K-quant dequant, GQA dequant, GGUF float16, vision, Mamba backends), so even a working build from the old tag would ship known-buggy code.
- Trimmed stale marketing copy in the snap summary/description ("120KB, 28 tok/s", "5 models") that no longer matches current reality.

## Verified
- Compiled and linked the corrected recipe locally against the current repo source tree: produces a real ELF binary (warnings only, no errors). Confirms the fix is mechanically correct, not just plausible-looking.

## Out of scope (flagging, not fixing here)
None of snap/AUR/homebrew have any CI automation actually publishing them anywhere — no `snapcraft push`/`brew`/AUR step exists in any workflow. This PR fixes the build; whether to invest in actually shipping these channels is a separate decision.


bong-water-water-bong:
Closing as superseded — same net diff as #595/#596 (the "clean" rebase), which is merging. This PR's two commits produce the identical end state as #596's single commit for these files.

PR #595, opened 2026-07-21
fix(packaging): snap/AUR/homebrew reference a deleted source file

bong-water-water-bong:


bong-water-water-bong:
Closing as superseded — #596 contains this exact commit (`d6475ced`) plus the numbers.json regeneration on top. Merging #596 instead.

PR #601, opened 2026-07-21
docs: update model catalog — 8 new 1BP model additions

bong-water-water-bong:

PR #602, opened 2026-07-21
fix: native 1BP export no longer overwrites HF PEFT adapter_config.json

bong-water-water-bong:


bong-water-water-bong:
Closing — same as #601: verified this branch's tree is byte-identical to `origin/main`; the 1BP export config-overwrite fix already landed on `main` via commit `ad003417`. Nothing left to merge.

Issue #54, opened 2026-07-12
Process gap: correctness/performance changes get merged without real hardware validation


bong-water-water-bong:
## Summary

A 2026-07-12 audit (see #47/#48/#50/#51/#52/#53, and the site/README correction commits from that day) found that nearly every falsified benchmark number and undetected bug traced back to the same root cause: changes affecting correctness or performance numbers get merged on "compiles cleanly" rather than "confirmed working on real hardware." This is a process gap, not a one-off mistake, and it's worth fixing at the root rather than continuing to catch instances of it after the fact.

## Evidence

- **`86350d86`** (RoPE convention, causal masking, quant-scale fix, applied across 19 engine variants): the commit's own message says *"Not yet validated against real NPU inference output ... representative files were also linked successfully."* It was merged anyway. The resulting numbers (97 tok/s for v12, and by extension whatever `all_5`/other variants inherited) sat in the README and site as fact for a day before anyone ran them on hardware.
- **`npu_engine_all.cpp`** (#52): a missing closing brace made the decode loop's batch-completion code run once per layer instead of once per batch. This shipped with the 2026-07-02 "28 tok/s, 5 models, 0 crashes" commit and was never caught, because nobody re-ran it after — the number was just carried forward into every subsequent doc/site update for 10 days.
- **Fused engine "coherent" claim**: README's main performance table said `✅ coherent` for NPU fused (291 tok/s) at the same time `site/benchmarks.json` — the repo's own stated "single source of truth" — said `raw — not yet coherent` for the identical engine. Nobody reconciled the two. The actual root cause (#44, lm_head/embedding weights silently zero-filled instead of dequantized) had already been diagnosed precisely, with a fix sketch, and was just sitting unfixed while the site kept the contradictory "coherent" claim live.
- **PR #42** (fused-engine tokenizer/decode-loop fix): correctly fixed two real bugs, verified via a 5-token smoke test — but that smoke test only checked "produces varied token IDs," not "produces a working end-to-end generation." The PR merged; end-to-end generation is still broken (#44).

## The pattern

In every case: a fix that's real and necessary gets merged based on a partial check (compiles, links, produces *some* different output, or "not yet validated" stated openly in the commit itself) — and then the *conclusion* ("this benchmark is now X tok/s," "this engine is now coherent") gets published as if the partial check had confirmed it.

## Suggested fix

[...]

PR #675, opened 2026-07-22
jarvis: full Rust port (server, RAG, tools, planner, STT, TTS, voice cloning)

bong-water-water-bong:
## Description

Ports the entire jarvis local-agent backend (`jarvis/*.py`) to Rust as a new `jarvis-rs/` crate, matching this repo's existing `rust/` (`onebit`) crate conventions (axum/tokio/reqwest). The HTTP API surface is unchanged — same endpoints, same JSON shapes, same SSE framing — so 1bit Mobile and the `/chat` web UI need zero client-side changes.

Also includes a standalone fix: `jarvis/ui.py`'s `CHAT_HTML` was a non-raw Python string containing embedded JS with `\uXXXX`/`\n` escapes that Python's own parser was silently eating before they ever reached the browser — crashing every request (`UnicodeEncodeError`) and, once that was patched, breaking every button on the page (`SyntaxError` killing the whole `<script>` block). Ported to a Rust raw string, which structurally can't have this bug class.

### What moved to Rust
- `server.py` → HTTP routing, handlers, SSE streaming
- `routing.py` → NPU/unified/GPU backend proxying
- `tools.py` → tool-call loop (brace-matched parsing) + permission gate + audit log
- `planner.py` → multi-step decomposition + per-subtask routing + synthesis
- `rag.py` → filesystem keyword search + session memory
- `stt.py` → STT, swapping faster-whisper (CTranslate2) for whisper-rs (whisper.cpp) with the equivalent ggml "tiny" model
- `tts.py` → Piper fallback (still just shells out to the `piper` CLI, unchanged approach)
- `voice/engine.py` + `voice/codec.py`'s decode path → `candle-nn` port of the RVQ-VAE codec decoder
- `audio_out.py`, `beacon.py`, `ui.py` → speaker mirror, LAN beacon, chat UI

### What stayed Python
`voice/train.py`, `voice/dataset.py`, `voice/record.py`, and `voice/codec.py` (offline voice-clone data collection/training tooling, never imported by the running server; `dataset.py` still imports `codec.py` directly).

## Type of Change

- [x] Other (jarvis backend — local agent server, not the core NPU/GPU inference engine)
- [x] Documentation (`jarvis/README.md` updated to describe `jarvis-rs` as the running server)

## Verification

- [x] Built and ran locally — `cargo build` and `cargo clippy` clean
- [x] `/chat` page load + JS syntax validated (`node --check`)
- [x] `/v1/chat/completions` non-streaming and real incremental SSE streaming, verified live against a running Ollama-routed model
- [x] Tool-call loop verified live: safe tool auto-allowed, sensitive tool denied without `allow_write` and permitted with it, audit log entries match the Python format
- [x] Server-side multi-turn memory verified live: correct recall across requests with zero client-sent history
- [x] `/v1/knowledge/*` (list/search/upload, both JSON and multipart) verified live
- [x] `/v1/agent/plan` verified live, including its error-surfacing format when a backend is down
- [x] STT verified via round-trip: fed jarvis's own Piper-synthesized speech into `/v1/audio/transcriptions`, got recognizable text back
- [x] TTS (Piper path) verified live: real non-silent 22kHz WAV output, confirmed through the physical USB speaker mirror
- [x] LAN beacon verified via live UDP capture — correct `{"service":"1bit",...}` payload
- [ ] Voice-cloning codec decoder (`voice.rs`) is **not** verified against real trained weights — no `.voice` pack has ever been created on this machine (`voice/record.py` has never been run), matching the Python original's own dormant state for this feature. Implemented layer-for-layer against `codec.py`'s architecture; will need a real pack to verify numerically.
- [ ] Full 1bit Mobile phone round trip (pairing via beacon, live mic recording) not independently re-tested against this Rust server — same API surface already verified via curl, but a live phone test is pending.

## Related Issues

N/A — internal refactor for the AMD hackathon Track 2 submission (jarvis is the Track 2 entry's base).

---
*By submitting this PR, I confirm that any benchmark numbers or status changes accurately reflect real on-device measurements, or are explicitly marked `unvalidated` per the repo's process policy (see #54).*


bong-water-water-bong:
Feature request — not a bug

PR #720, opened 2026-07-22
feat(vl): lightweight image preprocessing for VL models (zero OpenCV dep) + vision preprocessor

bong-water-water-bong:
Merging VL preprocessing branch: HIP resize/normalize kernel, VlProcessor class, unified_server + vision_server integration. 1258 lines across 8 files.


bong-water-water-bong:

PR #724, opened 2026-07-22
feat: 1BP ternary (TQ2) format — loader + converter + format header

bong-water-water-bong:
1BP ternary format support: onebp_loader.cpp, gguf_to_onebp.py updates, format header. 137 insertions across 4 files.


bong-water-water-bong:

PR #726, opened 2026-07-22
docs: INT8 on NPU2 via MLIR-AIE is architecturally blocked, use BFP16 + backend unification under C++23 + test_backend CMake target

bong-water-water-bong:
Documentation and code: NPU INT8 MLIR-AIE architectural blocker, C++23 backend unification, test_backend CMake target, 5 bug fixes. 1230 insertions across 40 files.


bong-water-water-bong:

Issue #844, opened 2026-07-23
ZINC Vulkan shaders (embed.spv) are never built → dense models fell back to CPU


bong-water-water-bong:
## ✅ RESOLVED — dense GPU inference through ZINC is correct and enabled by default (PR #854)

The C++ ZINC Vulkan backend now runs dense GGUF models (ZR1/Qwen2) end-to-end on the GPU and matches `cpu_generic` **token-for-token**, at **26.3 tok/s vs 6.6 on CPU (~4x)**.

### Final root cause
`max_seq` came straight from `context_length` (**131072**) -> KV cache **7.5 GB**, V region starting at **~3.7 GB**, past RADV's `maxStorageBufferRange` (4 GiB). Cached-V reads silently returned zero -> attention collapsed. Fix: cap `max_seq` at 8192.

### Full list of bugs fixed to get here (all merged)
1. ZINC shaders never compiled (wrong CMake glob dir) + runtime path resolution (#847)
2. Q4_K quantizer scale /255 -> /15
3. bf16 scale decode read as int16
4. Missing pipeline barriers between batched dispatches
5. FFN residual base not refreshed
6. `embed.comp` filled only first 256 of hidden
7. Zero embeddings (upload skipped on row-count mismatch)
8. `dmmv_q4k` shared-mem OOB at inter=8960
9. Untied lm_head + F32 lm_head path; QKV bias (Qwen2)
10. F32 weight path (removed requant error)
11. 2D dispatch grid for gemv_f32 (lm_head M=vocab > 65535 workgroup limit)
12. **KV cache max_seq cap (this fix) — the one that made attention correct**

### Validation method (reusable)
Per-op bisection against `cpu_generic`: `ZINC_NUM_LAYERS`/`CPU_NUM_LAYERS`, `ZINC_DEBUG_OPS`/`CPU_DEBUG_OPS`, `debug_readback`. normed/QKV/bias/RoPE matched exactly; only flash_attn's cached-V read was zero -> pinpointed the buffer-range issue.

### Follow-ups (not blocking)
- Validate other dense architectures (Llama, Mistral) through ZINC; add qk-norm for Qwen3 if it's ever routed to ZINC (currently -> NPU).
- Optional: restore a *correct* Q4_K path (sub-block scales) to cut the F32 memory footprint.

Issue #843, opened 2026-07-23
Zamba2 via unified_server: passes weights dir instead of model file → segfault


bong-water-water-bong:
## Summary
Starting `unified_server -w <dir>` with a Zamba2 GGUF (e.g. `zamba2-1.2b-instruct-v2-q4_0.gguf`) crashes with a **segfault** during model load.

## Repro
```
./build/unified_server -w /dir/with/zamba2.gguf -p 8104
```
Server log:
```
Loaded GGUF-native tokenizer from .../zamba2-1.2b-instruct-v2-q4_0.gguf (BOS=1, EOS=2)
Zamba2: Loading model from /tmp/bd_8110        <-- the DIRECTORY, not the file
[zamba2] Failed to open GGUF: /tmp/bd_8110
Zamba2: Failed to load model
Segmentation fault (core dumped)
```

## Root cause
The Zamba2 backend is handed the **weights directory** instead of the discovered model's `model_path`, so `open GGUF` fails, and the subsequent code dereferences the null/failed model → segfault.

## Fix
1. Pass the resolved `ModelConfig.model_path` (the `.gguf` file) to the Zamba2 backend, not the weights dir.
2. After a failed load, return cleanly instead of dereferencing a null model (defensive null-check).

Found while benchmarking the Zyphra family end-to-end on Strix Halo (gfx1151).

bong-water-water-bong:
Verified in #886: `src/backend_zamba2.cpp`'s `init()` already prefers `cfg.model_path` (the discovered `.gguf` file) over `weights_path` (the directory), with a clean `return false` if neither is available — already fixed in this tree. Closing.

PR #845, opened 2026-07-23
Engine crash-hardening: backend failover + graceful ZINC skip (+ Zyphra on-device validation)

bong-water-water-bong:
## Summary
Engine crash-hardening + on-device validation of the Zyphra family. Found and fixed a real server-crash bug while benchmarking all six models on Strix Halo (gfx1151).

## 🔒 Crash-hardening (the important one)
Dense models were **crashing the entire `unified_server`**: the ZINC Vulkan backend throws `Cannot open shader: embed.spv` on a lazy shader load, and that throw escaped on the **PILOT prefetch worker thread** → `std::terminate`.

- `backend_zinc.cpp`: resolve the shader dir robustly (`ZINC_SHADER_DIR` env → compile define → candidate build/install paths) and **validate `embed.spv` exists in `init()`**, returning `false` gracefully *before* starting Vulkan/PILOT.
- `backend_manager.cpp`: wrap `init()`/`benchmark()`/`generate()` (primary + failover) in try/catch so any backend exception becomes a **failover, not a crash**.

**Verified on-device:** ZR1-1.5B now fails ZINC over to `cpu_generic` and returns coherent output instead of aborting the server.

## 📊 On-device validation (2026-07-23, Radeon 8060S / gfx1151)
- **BlackMamba-1.5B: 84.2 tok/s**, **2.8B: 48.5 tok/s** (Mamba1 HIP) — confirms published 79.8 / 46.4.
- **ZR1-1.5B**: runs end to end, coherent output.
- README Zyphra table gets an on-device validation footnote.

## 🐛 Bugs filed during the sweep
- #843 — Zamba2 via `unified_server` passes the weights *dir* instead of the model file → segfault.
- #844 — ZINC Vulkan shaders (`embed.spv`) are never built (CMake globs the wrong dir) → dense models fell back to CPU.


bong-water-water-bong:
Superseded by a clean branch (squash-merge divergence made this diff noisy). Reopening as fix/gpu-crash-hardening.

Issue #701, opened 2026-07-22
Medium: model-switch reload in unified_server.cpp blocks all concurrent traffic


bong-water-water-bong:
## Description

\`tools/unified_server.cpp:877-890\`: the \`/v1/chat/completions\` model-switch path performs a full synchronous \`mgr.init()\` reload while holding both global mutexes (see also the related, more severe issue about these mutexes being held for the entire generation loop).

A client alternating the \`"model"\` field between two valid discovered models on successive requests forces a full reload+backend-init every time, each one blocking all concurrent traffic server-wide for the reload duration.

## Fix

Move model reload/switch off the hot request path where possible (e.g. background pre-warming, or a narrower lock scoped only to the switch itself rather than the whole mutex-guarded generation flow).

## Found via

Independent audit of \`tools/unified_server.cpp\`, focused on availability/performance under realistic multi-model usage patterns. Verified against current file content (2026-07-22).

bong-water-water-bong:
Fixed.

Issue #696, opened 2026-07-22
CRITICAL: uncaught json::type_error in /v1/chat/completions message parsing crashes entire server


bong-water-water-bong:
## Description

\`tools/unified_server.cpp:791-856\`: the try/catch at lines 793-800 only guards \`json::parse(req.body)\`. The subsequent field-access code (building \`prompt\`/\`vision_images\` from \`body["messages"]\`) has no catch at all.

If any \`messages[]\` element isn't a JSON object — e.g. \`{"messages":["hi"]}\` — \`msg.value("role","user")\` throws \`json::type_error\`, uncaught, propagating out of the httplib worker-thread lambda. This calls \`std::terminate()\`, crashing the **entire server process** for every connected client, not just the one that sent the malformed request.

Distinct from the already-fixed old finding #1 in \`AUDIT_ISSUES.md\` (that was catch-everything-*silently*-swallowing-errors in \`tests/zaya_server.cpp\`; this is catch-*nothing*-at-all, on a different code path in \`tools/unified_server.cpp\`).

## Fix

Wrap the full request-body-to-prompt conversion in a try/catch that returns HTTP 400 on any parse/type error, not just the initial \`json::parse\` call.

## Found via

Independent audit of \`tools/unified_server.cpp\`, focused on crash-resistance against malformed requests. Verified against current file content (2026-07-22).

bong-water-water-bong:
Fixed.

PR #24, opened 2026-07-04
feat(tools): vendor video-lora as tools/video-lora

bong-water-water-bong:
## Summary
- Vendors the standalone [bong-water-water-bong/video-lora](https://github.com/bong-water-water-bong/video-lora) project (Wan2.2/LTX-Video/AnimateDiff/CogVideoX generation w/ LoRA support, plus a pure-Zig Vulkan compute backend) into this repo at `tools/video-lora/` via `git subtree` (squashed, source at commit `1f5449a6`).
- Fixes `vulkan/build.zig` for the current `std.Build` API (`createModule`/`root_module` — the old single-arg `linkSystemLibrary`/`default_step` form no longer compiles).
- Moves CI from the inert nested `tools/video-lora/.github/workflows/ci.yml` (GitHub only reads workflows at the repo root) to `.github/workflows/video-lora-ci.yml`, scoped with a `paths:` filter to `tools/video-lora/**`.
- References the new subdirectory from the root README's architecture map and from the vendored project's own README.

## Why
This project was previously living loose in the user's home directory as a sibling checkout, and a prior scaffolding mistake had leaked its `README.md`/`.gitignore`/`pyproject.toml`/`src/video_lora/` directly into the 1bit.systems repo root as uncommitted changes. This PR gives it a proper, permanent home as a subdirectory instead.

## Test plan
- [ ] `cd tools/video-lora && pip install -e ".[dev]" && pytest -v`
- [ ] `cd tools/video-lora/vulkan && zig build` (Zig 0.15.2+, needs `glslc` + Vulkan headers)
- [ ] Confirm `video-lora-ci.yml` triggers only on `tools/video-lora/**` changes

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---
## Summary by cubic
Vendors `bong-water-water-bong/video-lora` into `tools/video-lora/`, adding text-to-video generation with LoRA support (Wan2.2, LTX-Video, AnimateDiff, CogVideoX) and a pure-Zig Vulkan compute backend. Sets up scoped CI and resolves linting and `mypy --strict` errors so the project builds and tests in the monorepo.

- **New Features**
  - Adds `tools/video-lora/` with pipelines and CLI (`generate`, `list-models`)
  - Unified LoRA loader for HF repos and local `.safetensors`
  - Pure-Zig Vulkan backend under `tools/video-lora/vulkan/`
  - Updates root README to reference the new tool

- **Bug Fixes**
[...]

PR #22, opened 2026-07-04
[gpu] H2O KV cache eviction + RadixAttention + CPU offloading + dynamic quantization

bong-water-water-bong:
## Summary

Four KV cache efficiency techniques for long-context GPU inference, all composing via shared zero-page remapping. **31/31 unit tests pass.**

### H2O Eviction (core)
- `scheduler/kv_cache.zig`: Cumulative attention score tracking, min-heap eviction, 3 policies
- Zero-page technique: evicted pages → reserved zero page → ~0 attention, no shader changes
- `scheduler/scheduler.zig`: Page alloc on admit, free on release
- `compute/forward.zig`: `allocOrEvict()`, zero-page remap in `uploadActivePageTable()`, `flushH2OScores()`
- `compute/forward_metal.zig`: Paged attention config + eviction check

### Attention Score Feedback (Vulkan)
- `flash_attn.comp`: Binding 6 (score_accum), per-token softmax accumulation
- Pipeline layouts 6→7 bindings, SPIR-V rebuilt

### New modules
- `radix_tree.zig`: Prefix tree for cross-request KV page sharing
- `offload_engine.zig`: CPU buffer pool with async offload/reload
- `quant_profile.zig`: 8 schemes, per-layer quantization profile

### Config
```bash
ZINC_KV_EVICTION_POLICY=h2o ./engine
```

---
## Summary by cubic
Adds long‑context GPU inference with H2O KV‑cache eviction, RadixAttention sharing, CPU offload, dynamic quantization, and a fused NPU+GPU engine served by a zero‑dep C++23 daemon. Finalizes the Windows port via `platform.h`, switches all Vulkan pipelines (including flash attention) to wave32, rebuilds the video diffusion engine with CacheDiT and multi‑LoRA, and updates docs/benchmarks (gemma3 4B now 122 tok/s).

- **New Features**
  - H2O eviction with attention feedback: coherent per‑token `score_accum` buffer (binding 6), min‑heap eviction, zero‑page remap; stable 7‑binding layout across single/batched shaders.
  - RadixAttention: prefix tree for cross‑request KV page reuse.
  - CPU offload: cold KV pages swap to RAM with async reload; transparent to kernels.
  - Dynamic quantization: per‑layer KV schemes in page table bits; dequant on read.
[...]

PR #850, opened 2026-07-23
Repo hygiene, honest claims, and resolve all HIGH/MEDIUM audit findings

bong-water-water-bong:
Two commits. The first is docs/hygiene; the second is a real concurrency fix.

## 1. Repo hygiene + honest claims (`c01d08bc`)
- **Untrack 41 committed ELF build outputs** (`tools/*`, `engine/fusion/*`, `engine/npu/tokenizer/*`, and staged first-party binaries under `snap/`, `packaging/`). These were compiled outputs already listed in `.gitignore` but never removed from the index (`.gitignore` only affects untracked files). Sources are untouched; `git rm --cached` only.
- **`.gitignore`**: replaced the brittle hand-enumerated tool list with rules that keep sources/scripts and reliably drop extensionless build artifacts.
- **README**: the "One Binary" subtitle/lead now describe reality — one shared in-process engine behind thin `zaya_server`/`unified_server`/`onebit`/`onebitd` binaries, not a single ~400 KB binary doing all three.
- **README**: Tests badge now tracks live CI status instead of a static `16/17` that conflicted with `AUDIT_ISSUES.md`'s `10/13`.
- **`validate-claims.yml`**: header no longer claims daily on-hardware re-measurement — the driver scripts are retired and the steps self-skip, so the job only checks `latest.json` internal consistency.
- `CLAUDE.md` → `AGENTS.md` symlink (were byte-identical; prevents drift).

## 2. Fix HIGH-severity data race — AUDIT_ISSUES.md #2 (`5086f017`)
In `unified_server.cpp`, the actual decode (`generate_completion` → `mgr.reset`/`mgr.generate`), the model-switch `mgr.init()`, and vision `forward_embed()` ran with **no lock held** — the earlier `#696` change released the lock around the hot path to chase concurrency a single-context backend can't provide. Two concurrent `/v1/chat/completions` or `/v1/completions` requests raced on the one shared KV cache / active-backend pointer (corruption / segfault under load). `/v1/backend/select` guarded the same `mgr` mutation with `g_config_mutex`, which the decode path no longer held — so its "same mutex" comment described an already-broken invariant.

**Fix:** a dedicated, outermost `g_inference_mutex` serializes all shared-backend compute (decode, model reload, backend switch). Existing `g_config_mutex`/`g_strategy_mutex` nest inside it; `/v1/health` and `/v1/models` still take only config+strategy, so health stays responsive during a decode (preserving the `#701` goal). `zaya_server.cpp`'s half of #2 was already fixed behind `g_router_mutex`.

**Validation** (ROCm hardware not available in this environment):
- Full object compile of the TU is clean (`g++ -std=c++23 -c`).
- A TSan + UBSan model of the exact lock topology (8× each handler concurrently) shows no deadlock, no overlapping decode (`assert(n==1)` holds), and no data race across repeated runs.
- Lock order audited: every `g_inference_mutex` acquisition is outermost; no path takes config/strategy then inference.

**Known residual (low, documented in AUDIT_ISSUES.md):** `/v1/backend/status` and `/v1/models` read `mgr.active_info()` under config+strategy only — a benign stale-pointer read; fully closing it wants an atomic active-backend pointer inside `BackendManager`.

Rebased on latest `main` (on top of #849); no file overlap with upstream.


bong-water-water-bong:
Added a third commit: **fix AUDIT #3 — NPU worker SIGPIPE crash** (`7c4c8755`).

Most of #3 was already fixed and the audit text was stale — `spawn()` already does a `READY\n` handshake (sets `ready` only on success), `shutdown()` already escalates quit → SIGTERM → SIGKILL with `waitpid` reaping at each step, and the double-pipe path already avoids the FD leak. `backend_flm.cpp` no longer exists.

The one real remaining bug: once the worker dies, the `write()` calls in `gemm()`/`shutdown()` hit a reader-less pipe, and **SIGPIPE's default action terminates the whole server**. Only `token_router.cpp` ignored it; `unified_server`/`zaya_server` did not. `spawn()` now ignores SIGPIPE process-wide once (thread-safe local static), so those writes return `EPIPE` and are caught by the existing short-write checks.

Verified with a standalone repro: default disposition = killed-by-signal-13; with the ignore, `write()` returns `EPIPE` and the process survives. `AUDIT_ISSUES.md` #3 marked resolved with a noted low-severity D-state zombie residual.

bong-water-water-bong:
Added a fourth commit: **harden all remaining audit items** (`678bb12a`).

I re-verified **every** item in `AUDIT_ISSUES.md` against the current tree — the 2026-07-18 snapshot is heavily stale. **All HIGH and MEDIUM findings are resolved.** Most were already fixed in prior work (#5, #6, #7, #8, #9, #12, #14, #16, #17, #18, #19, #20, #21, #23, #28) or moot (#25 — no `1bit/` sub-monorepo; #27 — zero Python files). A full status matrix for all 28 is now at the top of `AUDIT_ISSUES.md`.

Genuinely-open fixes in this commit:
- **#4 (HIGH, DoS)** — `gguf_zamba2_loader.cpp` trusted 64-bit string/array lengths from the model file (`std::string(len)` → a crafted GGUF OOMs the process). Capped to 1 MiB / 1M elements, matching `gguf_reader.cpp`; oversized fields are skipped in-sync or refused. (The other readers were already capped; `tokenizer` uses a `uint16_t` length, inherently ≤64 KB.)
- **#10 (MED)** — `set -euo pipefail` added to 34 executable scripts; scripts meant to be `source`d (`env.sh`, `toolchain-env.sh`, `jarvis-world-sweep.sh`) excluded so they don't poison the caller's shell. All re-checked with `bash -n`.
- **#11 (MED)** — prod paths already resolve `$HOME`/XDG; fixed the two real runtime hardcodes (`build_06b_kernels.sh` default → `$HOME`, `npu_engine_v12.cpp` model path → `NPU_MODEL_PATH`). Remaining `/home/bcloud` hits are in comments.

Deferred (LOW/cosmetic, with rationale in the doc): **#13** (header→cpp move, recompile-only, regression risk), **#15** (C-style-cast sweep, style-only, high churn), **#22** (`warpSize` `static_assert` not reliably valid on HIP host), **#24** (TODO→issue tracking).

Validated: changed C++ TUs pass `g++ -fsyntax-only`; all shell scripts pass `bash -n`. Note: `-u`/nounset may surface latent unset-var use in scripts I couldn't execute (no hardware) — that's intended hardening.

bong-water-water-bong:
Closing in favor of #892 — this PR's base was stale enough that a direct merge/update would've pulled in a large amount of unrelated historical cruft (old tracked build artifacts, etc.). #892 cherry-picks the same 5 commits cleanly onto current `main`, verified as genuinely new (not superseded, unlike the 4 other open PRs closed alongside this one), full build + ctest green.

PR #862, opened 2026-07-24
fix(tests): resolve test_pipeline_real segfault — closes #825

### LINKED ISSUE #825: CRITICAL: test_pipeline_real SEGFAULT — UB from INFINITY with fast-math + ignored HIP error returns
## Summary
`test_pipeline_real` crashes with SEGFAULT every run (test #16/17 fails). Root cause is a combination of undefined behavior and unchecked errors.

## Bug 1: UB — INFINITY with fast-math enabled
Multiple lines in `engine/fusion/zero_copy/test_pipeline_real.cpp` use `INFINITY`/infinity with fast-math floating-point options enabled, which is **undefined behavior**:
- Lines 83, 85, 124, 129, 140, 150, 267
- The compiler warns: `use of infinity is undefined behavior due to the currently enabled floating-point options [-Wnan-infinity-disabled]`

## Bug 2: Ignored `[[nodiscard]]` HIP error returns
7 calls to HIP functions return `hipError_t` marked `[[nodiscard]]` but the return value is cast to `(void)`:
- Lines 168, 206, 207, 212, 213, 239, 293
- If `hipMalloc` fails (OOM), `d_K/V/Q/AttnOut` remain uninitialized -> null pointer dereference -> SEGFAULT

## Severity
**CRITICAL** — blocks 1/17 tests from passing (SEGFAULT)

bong-water-water-bong:
### Bug 1: `-ffast-math` was silently defeating all `std::isfinite()` guards
`-ffast-math` includes `-ffinite-math-only`, which tells the compiler NaN/Inf never occur — so every `std::isfinite()` check became a dead-code no-op.

**Fix:** Replaced with specific flags that give the same perf without suppressing NaN/Inf:
```
-fno-signed-zeros -fno-trapping-math -fassociative-math
-freciprocal-math -fno-math-errno -fno-rounding-math
```
(Notably **not** `-ffinite-math-only` or `-funsafe-math-optimizations`)

### Bug 2: NULL rmsnorm weight pointer dereference
`npu_ffn` called `rmsnorm_f32(h, nullptr, H)` → `w[i]` dereferenced nullptr → SEGFAULT.

**Fix:** Added `w ? w[i] : 1.0f` guard inside `rmsnorm_f32`.

### Bug 3: Six unchecked `[[nodiscard]]` HIP calls
| Call | Before | After |
|------|--------|-------|
| `hipSetDevice(0)` | unchecked | ✅ |
| `hipStreamCreate()` | unchecked | ✅ |
| `hipMemsetAsync()` ×4 | unchecked | ✅ |
| `hipStreamSynchronize()` | unchecked | ✅ |
| `rcpp_kv_cache_attn_decode()` | unchecked | ✅ |
| `hipStreamDestroy()` | unchecked | ✅ |

### Bug 4
`(void)(void)hipFree` typo — already fixed in a prior pass.

---

**Verification:** Both files compile clean with the updated flag set. No behavioral change to non-NaN/Inf inputs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -883,7 +883,7 @@ if(hip_FOUND AND xrt_FOUND AND aiebu_FOUND)
             ${CMAKE_SOURCE_DIR}/include
             ${THEROCK_SDK_BASE}/include)
         target_compile_options(test_pipeline_real PRIVATE
-            -O3 -march=native -ffast-math -std=c++23 -fopenmp -D__HIP_PLATFORM_AMD__)
+            -O3 -march=native -fno-signed-zeros -fno-trapping-math -fassociative-math -freciprocal-math -fno-math-errno -fno-rounding-math -std=c++23 -fopenmp -D__HIP_PLATFORM_AMD__)
         target_link_libraries(test_pipeline_real PRIVATE
             xrt_coreutil xrt_core uuid rt pthread
             /usr/lib/x86_64-linux-gnu/libaiebu.a)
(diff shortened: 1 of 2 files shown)


bong-water-water-bong:
Closing as superseded — this fixes the exact same bug (#825) I already fixed and merged via #886 earlier this session, just with slightly different code (this PR's inline `w ? w[i] : 1.0f` ternary vs my `if(w){...}else{...}` branch, and equivalent HIP-return-value checks). Confirmed via a local test-merge: the only real conflict was in `rmsnorm_f32`/the HIP error-check sites, both already fixed on `main`. Nothing unique left to bring in.

Issue #1010, opened 2026-07-26
zaya_persistent_moe.hip: unbounded spin-wait in Phase-2 handoff will deadlock the GPU once wired up (currently latent)


bong-water-water-bong:
## Summary

`kernels/zaya_persistent_moe.hip:210`:
```cpp
while (load_flag(done_flags, layer, 1) == 0) { __threadfence_block(); }
```
has no timeout. The same file's Phase-1 router wait (lines 128-137) was given a `MAX_SPINS`-bounded timeout specifically to fix a past GPU hang (referenced in-file as "issue #405"), but the equivalent Phase-2 wait for the gate_up→down handoff was left unbounded.

`persistent_moe_kernel` launches `GRID_X*2*N_LAYERS = 20480` blocks against a device that can only run ~1280 waves concurrently (per the file's own comment). A down-block scheduled before its layer's gate_up blocks complete can spin forever, starving the wave slots those gate_up blocks need to make progress — a classic persistent-kernel deadlock.

## Status

Verified via grep that `persistent_moe_kernel` is not currently launched anywhere in the codebase (not wired into `zaya_engine.cpp` despite the file being `#include`d) — this is currently latent, not live. Filing so it's fixed before anyone wires this kernel up and hits exactly the hang class issue #405 was meant to close off everywhere.

## Suggested fix

Apply the same `MAX_SPINS`-bounded timeout used in the Phase-1 wait to the Phase-2 wait at line 210.

bong-water-water-bong:
Fixed in #1011 (merged): applied the same `MAX_SPINS`-bounded timeout used for the Phase 1 wait (issue #405) to the Phase 2 gate_up→down handoff wait.

Issue #1009, opened 2026-07-26
tools/run_trace_validate.sh: TOCTOU symlink race on predictable /tmp path before sudo invocation


bong-water-water-bong:
## Summary

`tools/run_trace_validate.sh:21-22,38,64,67` uses a fixed, predictable path:
```bash
NPU_DIR="/tmp/trace_npu"
```
It runs `rm -rf "$NPU_DIR"` as the invoking user, then shortly after runs `sudo ./npu_engine_trace ... --trace "$NPU_DIR"` as root.

## Impact

A local attacker on a shared machine can win the race by recreating `/tmp/trace_npu` as a symlink between the `rm -rf` and the `sudo` invocation, potentially causing the root-privileged trace writer to create/overwrite files through the symlink (classic TOCTOU privilege-escalation pattern). Lower real-world severity since this is a maintainer-only dev tool, not shipped/user-facing, but it's a genuine, reproducible pattern.

## Suggested fix

Use `mktemp -d` to create a non-predictable directory instead of a fixed path under `/tmp`.

bong-water-water-bong:
Fixed in #1011 (merged): `NPU_DIR` now created via `mktemp -d` right before the sudo-privileged write instead of a fixed, predictable `/tmp` path.

Issue #1008, opened 2026-07-26
packaging/model-download.sh + packaging/aur/PKGBUILD: no integrity verification of downloaded artifacts


bong-water-water-bong:
## Summary

**1. `packaging/model-download.sh:22-28,79`:**
The `MODELS` registry format is documented as `name|description|size|url|sha256`, but every entry's sha256 field is empty, and `download_model()` never checks it — `curl -fL ... -o "$OUTFILE"` with no hash verification at all.

**2. `packaging/aur/PKGBUILD:10`:**
```
sha256sums=('SKIP')
```
disables source-tarball checksum verification entirely. `makepkg` will build whatever it downloads from the GitHub archive URL with zero integrity check. Standard AUR guidance flags `SKIP` as inappropriate for tagged-release sources (it's meant for pure VCS/git-source packages only).

## Impact

Multi-GB model weight files and the release source tarball are both fetched over the network with no integrity verification — silent corruption or a compromised/mismatched upstream asset would go completely undetected.

## Suggested fix

- Populate real sha256 values in the `MODELS` registry and verify after download in `model-download.sh`.
- Pin an actual `sha256sums` value in `PKGBUILD` for the tagged release tarball.

bong-water-water-bong:
Fixed in #1011 (merged): `model-download.sh` now verifies sha256 when populated in the registry (and warns loudly when it isn't — the registry itself still needs real hashes filled in as a follow-up); `aur/PKGBUILD` pinned to the real v1.0.0 tarball hash instead of `SKIP`.

Issue #1007, opened 2026-07-26
packaging/mobile.sh + packaging/npu-install.sh: unbound variable crash under set -u on the common (no-ngrok / bash) path


bong-water-water-bong:
## Summary

**1. `packaging/mobile.sh:69,84,97` (identical copy served live at `site/mobile.sh`):**
`NGROK_URL` is only assigned inside the `if command -v ngrok` branch. When ngrok is not installed, the script falls into the `else` branch and later hits `if [ -n "$NGROK_URL" ]` at line 69 with the variable never set. Under `set -euo pipefail` this is a fatal "unbound variable" error. Reproduced by running the exact pattern in bash.

**2. `packaging/npu-install.sh:85`:**
```bash
[ -n "$ZSH_VERSION" ] && SHELL_CONFIG="${HOME}/.zshrc"
```
Documented usage is `curl -sL ... | bash`, and `ZSH_VERSION` is never set under bash. Reproduced: this line alone aborts a `set -u` script with "ZSH_VERSION: unbound variable".

## Impact

- `mobile.sh`: the documented `curl -fsSL https://1bit.systems/mobile.sh | sh` invocation aborts immediately after the "ngrok not installed" warning on any machine without ngrok pre-installed (the common case) — never reaches the Step 3/4 instructions.
- `npu-install.sh`: aborts right at the PATH-setup step, **after** models/binaries were already successfully installed, leaving the user with a half-finished install and a confusing error.

## Suggested fix

Initialize both variables up front: `NGROK_URL=""` in `mobile.sh`, and use `${ZSH_VERSION:-}` in `npu-install.sh`.

bong-water-water-bong:
Fixed in #1011 (merged): `NGROK_URL=""` initialized up front in both `packaging/mobile.sh` and `site/mobile.sh`; `${ZSH_VERSION:-}` in `packaging/npu-install.sh`.

PR #1011, opened 2026-07-26
fix: bug sweep cycle 3 — SSRF, XSS, hardcoded MoE dims, systemd hardening, more (#1001-#1010)

bong-water-water-bong:
## Summary

Parallel GitNexus-guided bug hunt across four areas of the codebase not covered by the previous sweep (`jarvis/rust/workers`, `spec-decode/ck-prefill/kernels`, `integrations/adapters/experimental`, `site/packaging/scripts`), followed by fixes for every confirmed bug. Full build + test verification before and after.

## Fixes

- **#1001** (highest severity — correctness): `zaya_moe_expert_ffn.hip` / `zaya_moe_batch_union.hip` had `H=2048`/`N_EXP=16`/`N_FF=2048` baked in as compile-time macros in the **live GPU MoE decode path**, silently corrupting inference for any model with different dimensions — directly contradicts this engine's model-agnostic goal. Threaded `h_dim`/`n_ff`/`n_exp` through as runtime kernel params (found and fixed 3 more leaked-macro pointer-arithmetic sites beyond the one originally scoped, in `batched_moe_union_kernel`, `batched_moe_fused_kernel`, and `moe_sorted_expert_kernel`).
- **#1002** (security): stored XSS in `site/analytics/index.html` + `site/s/index.html` — GitHub stargazer profile fields (bio/name/company) were interpolated into `innerHTML` unescaped. Added an `esc()` helper, applied to every untrusted field.
- **#1003**: `workers/auth-url` — `env` was never passed into `handleRequest`, so `VERIFICATION_CODES` was permanently unreachable; removed the orphaned `VERIFICATION_FILES`; added the missing `bing*.html` route; dropped unnecessary OAuth code/state logging.
- **#1004** (security): `integrations/comfyui` — SSRF via unrestricted `server_url` widget (now loopback-only by default, opt-in via `ONEBIT_ALLOWED_SERVER_HOSTS`), capped decoded image size against decompression-bomb responses, fixed pixel-value wraparound on the tensor→image cast.
- **#1005** (security): `npu-companion.service` ran as root with zero sandboxing — added a dedicated service user + standard systemd hardening directives (`NoNewPrivileges`, `ProtectSystem=strict`, etc).
- **#1006**: `site/_headers` CSP `img-src` was missing `avatars.githubusercontent.com`, silently breaking every stargazer avatar image.
- **#1007**: `mobile.sh` / `npu-install.sh` crashed under `set -u` on the common path (no ngrok installed / plain bash) — unbound `NGROK_URL` / `ZSH_VERSION`.
- **#1008** (security): `model-download.sh` never checked sha256 (registry had none populated); `aur/PKGBUILD` used `sha256sums=('SKIP')`. Added verification logic + pinned the real v1.0.0 tarball hash.
- **#1009** (security): `run_trace_validate.sh` had a TOCTOU symlink race between an unprivileged `rm -rf` and a `sudo`-privileged write through the same fixed `/tmp` path. Switched to `mktemp -d`.
- **#1010**: `zaya_persistent_moe.hip`'s Phase-2 gate_up→down handoff spun forever with no timeout (currently dead/unwired code, but a live GPU-hang risk the moment it's enabled) — applied the same bounded-timeout pattern already used for the Phase-1 wait (issue #405).

## Test plan

- [x] Full clean rebuild (346/346 targets) before starting
- [x] `ctest` 19/19 passing (3 skipped, missing hardware model files) before starting
- [x] Rebuilt after the #1001 MoE kernel fix — clean, 19/19 tests still pass
- [x] Rebuilt after the #1010 kernel fix — clean
- [x] Syntax-checked all shell scripts (`bash -n`), the Python file (`ast.parse`), and all JS (`node --check`, including inline `<script>` blocks)
- [x] `gitnexus detect_changes` — risk: low, no affected execution flows
- [x] Verified this branch's staged diff contains *only* these 18 files (repo has unrelated concurrent in-progress changes from another process that must not be swept into this PR)

### DIFF
--- a/integrations/comfyui/__init__.py
+++ b/integrations/comfyui/__init__.py
@@ -22,6 +22,7 @@ import json
 import os
 import base64
 import io
+from urllib.parse import urlparse
 import httpx
 from PIL import Image
 import numpy as np
@@ -31,6 +32,32 @@ import torch
 DEFAULT_UNIFIED_URL = "http://127.0.0.1:8088/v1"
 DEFAULT_IMAGE_URL   = "http://127.0.0.1:8089/v1"
 DEFAULT_AUDIO_URL   = "http://127.0.0.1:8090/v1"
+
+# ComfyUI's threat model is untrusted shared workflow JSON — a workflow can set
+# server_url to an arbitrary value. Without an allowlist that's an SSRF
+# primitive (internal metadata endpoints, internal admin APIs, etc). Default
+# to loopback-only; operators who genuinely run the 1bit server on another
+# host can opt in via ONEBIT_ALLOWED_SERVER_HOSTS (comma-separated hostnames).
+# See #1004.
+_LOOPBACK_HOSTS = {"127.0.0.1", "localhost", "::1"}
+_EXTRA_ALLOWED_HOSTS = {
+    h.strip() for h in os.environ.get("ONEBIT_ALLOWED_SERVER_HOSTS", "").split(",") if h.strip()
+}
+MAX_DECODED_IMAGE_BYTES = 64 * 1024 * 1024  # 64 MB cap on server-returned images
+
+def validate_server_url(server_url: str) -> str:
+    """Reject server_url values that aren't loopback / operator-allowlisted (SSRF guard)."""
+    parsed = urlparse(server_url)
+    if parsed.scheme not in ("http", "https"):
+        raise ValueError(f"server_url must be http(s): {server_url!r}")
+    host = (parsed.hostname or "").lower()
+    if host not in _LOOPBACK_HOSTS and host not in _EXTRA_ALLOWED_HOSTS:
+        raise ValueError(
+            f"server_url host {host!r} is not allowed. Only loopback hosts "
+            f"({sorted(_LOOPBACK_HOSTS)}) are permitted by default — set "
+            f"ONEBIT_ALLOWED_SERVER_HOSTS to allow other hosts."
+        )
+    return server_url
 
 # ─── Helper: image to base64 ──────────────────────────────────────
 def pil_to_base64(img: Image.Image, format: str = "PNG") -> str:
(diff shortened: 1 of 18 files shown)


bong-water-water-bong:
## CI status

Both **required** checks pass: `C++ (cmake configure + build)` ✅ and `Lint (clang-format, ruff)` ✅. `smoke-test` and `github-advanced-security` are not in the branch ruleset's required-checks list, but noting their failures here so they don't cause confusion:

**`smoke-test` — environmental, not a regression from this diff.** It starts a fresh `unified_server` on port 8099 and polls `/v1/health` for 40s. In both the original run and a rerun, the new server process starts, loads the tokenizer and allocates the KV cache, then never logs anything further and the health check times out (`curl` exit 7 — connection refused). On the runner itself there's a pre-existing, unrelated `unified_server` process that's been running since well before this PR (`pid 2618`, `-p 8088`, ~11GB RSS, started 10:53) holding the GPU/NPU. The new instance almost certainly can't acquire the hardware while that's alive. This matches this repo's known "unified_server is a singleton" issue — the workflow's `rm -f /tmp/unified_server.lock` only clears a stale lock *file*, it can't preempt a different already-running process that's actively holding the device. I re-ran the job once to rule out a one-off flake; it failed the same way both times, so this isn't going to self-resolve — it needs either that stray process cleaned up or the smoke-test job made robust to a busy GPU (e.g. failing fast with a clear message instead of a silent 40s timeout). Not something in this PR's 18 files.

**`github-advanced-security` — GitHub-side infra error, unrelated to this repo.** The job's own log shows `SessionModelError: 400 model_not_supported` from GitHub's Copilot backend (`claude-opus-4.6` rejected as unsupported) before it ever analyzes the diff. Not rerunnable via `gh run rerun` (not a repo workflow run).

I did not touch the stray `pid 2618` process or merge this PR — flagging both here for a human to decide on, since neither is reversible/low-stakes enough for me to act on unilaterally on a shared machine.

Issue #1043, opened 2026-07-27
187 git-tracked files are symlinks to absolute host-machine paths outside the repo — breaks on any clone but this exact box


bong-water-water-bong:
## What's wrong

Found while auditing the NPU xclbin duplication bugs (#1040, #1042). Checked whether any git-tracked files are missing/dangling on this machine — zero are, which is itself suspicious for a repo this size, so I checked *why*: 187 tracked files are actually symlinks pointing to **absolute paths on this specific machine's home directory**, entirely outside the git repo and not tracked by version control at all.

\`\`\`
git ls-files -z | while IFS= read -r -d '' f; do
  [ -L "$f" ] && case "$(readlink "$f")" in /*) echo "$f -> $(readlink "$f")";; esac
done
\`\`\`

## Breakdown (187 total)

| Target root | Count | What |
|---|---|---|
| \`/home/bcloud/zinc\` | 164 | Compiled Vulkan shader binaries (\`engine/fusion/shaders/vulkan/*.spv\`) |
| \`/home/bcloud/npu-sandbox\` | 15 | NPU xclbins (\`engine/npu/xclbins/*\`) — same files implicated in #1040/#1042 |
| \`/home/bcloud/fastflowlm-build\` | 6 | Vendor NPU xclbins (\`engine/npu/xclbins/*\`) — the "real" fix source used in #1042 |
| \`/home/bcloud/1bit-systems\` (self) | 1 | \`engine/fusion/npu_engine_universal\` → its own build output, same repo |
| \`/home/bcloud/.config\` | 1 | \`engine/fusion/model.q4nx\` → \`~/.config/flm/models/Qwen3-0.6B-NPU2/model.q4nx\` |

Full \`npu-sandbox\` and \`fastflowlm-build\` lists (the 21 most directly relevant to the NPU work):

\`\`\`
engine/npu/xclbins/final_i8_D_v.xclbin        -> ~/npu-sandbox/npu-infer/build/int8/final_i8_D_v.xclbin
engine/npu/xclbins/final_i8_GU_v.xclbin       -> ~/npu-sandbox/npu-infer/build/int8/final_i8_GU_v.xclbin
engine/npu/xclbins/final_i8_O_v.xclbin        -> ~/npu-sandbox/npu-infer/build/int8/final_i8_O_v.xclbin
[...]

Issue #1042, opened 2026-07-27
engine/npu/xclbins: gemma4_e2b's GU/D and (2 of) llama/qwen3_8b/qwen3_vl_4b's G/U kernels are wrong-shape clones — LIVE models, not dead files


bong-water-water-bong:
## Summary

Follow-up to #1040 (which covered 3 *orphaned, unwired* model tags). This is the more serious sibling: the same root cause also hit models that **are** wired into the live engine config (\`engine/npu/src/npu_dims.h\`).

## Evidence

\`\`\`
$ md5sum engine/npu/xclbins/final_i8_GU_*.xclbin
eea4aeab6181b3a410640b221a1c153f  final_i8_GU_gemma4_e2b.xclbin
eea4aeab6181b3a410640b221a1c153f  final_i8_GU_phi4.xclbin           (orphaned, see #1040)
eea4aeab6181b3a410640b221a1c153f  final_i8_GU_qwen3_1_7b.xclbin     (orphaned, see #1040)
eea4aeab6181b3a410640b221a1c153f  final_i8_GU_qwen3_4b.xclbin       (orphaned, see #1040)

$ md5sum engine/npu/xclbins/final_i8_D_*.xclbin
531a1cd0bb99885112ecee01ea9bcbf4  final_i8_D_gemma4_e2b.xclbin
531a1cd0bb99885112ecee01ea9bcbf4  final_i8_D_phi4.xclbin            (orphaned)
531a1cd0bb99885112ecee01ea9bcbf4  final_i8_D_qwen3_1_7b.xclbin      (orphaned)
531a1cd0bb99885112ecee01ea9bcbf4  final_i8_D_qwen3_4b.xclbin        (orphaned)

$ md5sum engine/npu/xclbins/final_i8_G_*.xclbin engine/npu/xclbins/final_i8_U_*.xclbin
845b2963923d3dc367deb6a52bf9f61d  final_i8_G_llama.xclbin
845b2963923d3dc367deb6a52bf9f61d  final_i8_G_qwen3_8b.xclbin
845b2963923d3dc367deb6a52bf9f61d  final_i8_G_qwen3_vl_4b.xclbin
207ecafdf19f2fcac93ba04bcb51100f  final_i8_U_llama.xclbin
207ecafdf19f2fcac93ba04bcb51100f  final_i8_U_qwen3_8b.xclbin
[...]

Issue #1040, opened 2026-07-27
engine/npu/xclbins: several QKV xclbins are byte-identical clones across unrelated models, and the models aren't wired into the engine at all


bong-water-water-bong:
## What's wrong

Found while digging into #1029 (using the actual torch2aie/MLIR-AIE toolchain, now confirmed available on this box). Checked whether \`engine/npu/xclbins/final_i8_QKV_*.xclbin\` files were genuinely per-model-compiled:

\`\`\`
$ md5sum engine/npu/xclbins/final_i8_QKV_*.xclbin | sort
009e83f2566808a1914252c58f1eec1f  final_i8_QKV_qwen3_0_6b.xclbin
06e5e004ef182769cf213760110d114d  final_i8_QKV_qwen3_vl_4b.xclbin
3d86f896e5c9b6b9cad5d1f08f200195  final_i8_QKV_phi4.xclbin
3d86f896e5c9b6b9cad5d1f08f200195  final_i8_QKV_qwen3_1_7b.xclbin
3d86f896e5c9b6b9cad5d1f08f200195  final_i8_QKV_qwen3_4b.xclbin
656134b07f6d5ce758e0b598b3225730  final_i8_QKV_qwen3_8b.xclbin
94cd4aa416465ad0a3271f22b1eef965  final_i8_QKV_llama.xclbin
ba5bdca45a6c8f4e6b3145c5a1279e82  final_i8_QKV_v.xclbin
d8f06dd664552a55b969c888cce6598b  final_i8_QKV_gemma4_e2b.xclbin
\`\`\`

\`final_i8_QKV_phi4.xclbin\`, \`final_i8_QKV_qwen3_1_7b.xclbin\`, and \`final_i8_QKV_qwen3_4b.xclbin\` are **byte-for-byte identical** despite Phi-4, Qwen3-1.7B, and Qwen3-4B having different hidden sizes/head configs (and therefore different QKV GEMM shapes) — these cannot all be genuinely correct, shape-specific compiled kernels.

## Root cause found

[...]

Issue #1036, opened 2026-07-27
End-to-End Smoke Test workflow has never passed — 15/15 historical runs failed, likely GPU contention with the always-on production unified_server


bong-water-water-bong:
## What's wrong

\`.github/workflows/end-to-end-smoke.yml\` (\`smoke-test\` job, runs on \`[self-hosted, strix-halo]\` — this exact box) starts its own \`unified_server --port 8099\` with Llama 3.2 1B Instruct, waits up to 40s for \`/v1/health\`, then hits chat/completions/timeout endpoints.

Checked its run history — **15/15 recorded runs have failed**, across completely unrelated PRs (\`perf/npu-v12-optimizations\`, \`fix/bug-sweep-2-1001-1010\`, this session's \`fix/backend-cpu-q4-quality-measurement\`). This is not tied to any specific change; it appears to never pass.

## Evidence from the latest failure

The job's own log only shows the \"Test health endpoint\" step failing with \`curl -sf ... exit code 7\` (couldn't connect) — the server never came up within the 40s wait window.

The redirected server output (\`/tmp/smoke_server.log\`, still present on this box after the run since the runner and this shell share the same filesystem) shows:

\`\`\`
Loaded BPE tokenizer from models//tokenizer.htok (BOS=1, EOS=0)
  KV cache: linear contiguous 4096 tok x 40 layers = 80.0 MB
\`\`\`

...and nothing further. The process got through tokenizer load and KV-cache allocation, then never printed anything else (no backend-init lines, no \"server listening\") before the job's 10-minute timeout killed it.

## Hypothesis (not yet confirmed)

This exact machine already runs a **persistent production \`unified_server\`** as a systemd-managed singleton (port 8088, currently pid 2961 when checked). \`tools/unified_server.cpp\`'s \`acquire_singleton_lock()\` flocks a single global path (\`$XDG_RUNTIME_DIR/unified_server.lock\` or \`/tmp/unified_server.lock\`) — not scoped per-port. The workflow's \`rm -f /tmp/unified_server.lock\` before starting only clears the \`/tmp\` path; if the production instance is actually using \`$XDG_RUNTIME_DIR/unified_server.lock\` (likely, since it runs as a user session/systemd-user service), that \`rm\` never touches the real lock, and/or the two instances contend for the same physical GPU (Radeon 8060S) regardless of the lock outcome, causing the CI instance to hang instead of failing fast.

[...]

PR #1047, opened 2026-07-27
fix: NPU BO sync, single-buffer layerB, NPU coexistence, CI fixes (#1029)

### LINKED ISSUE #1029: npu_engine_universal: per-layer NPU kernel execution extremely slow or hung (>3min for layer 1 of 28)
## What's wrong

With the \`.1bp\` data-correctness bugs fixed (see #1023 — offset double-counting, missing norm weights), \`npu_engine_universal\` now runs real inference instead of crashing/reading garbage. But actual per-layer execution is far slower than expected:

\`\`\`
$ timeout 180 ./engine/npu/build/npu_engine_universal models/Qwen3-0.6B.1bp 5
...
Dequant+pack...
  5537ms

=== Prefill 9 ===
  L0qaogd
  L1
[killed by 180s timeout, still on layer 1 of 28]
```

Dequant+pack (real work, all 310 tensors) completes in ~5.5s — reasonable. But actual NPU layer execution (the \`q\`/\`a\`/\`o\`/\`g\`/\`d\` sub-stage markers per layer) barely gets through layer 0 and into layer 1 in 3 minutes. The project's own documented numbers put NPU v12 at ~69 tok/s for this exact model — 28 layers should take low milliseconds each for a 5-token prefill+decode, not minutes.

## Not yet investigated

Could be:
- A genuine hang in the NPU kernel launch/wait sequence (XRT run.wait() never returning, bad kernel args, etc.)
- Extreme but real per-call overhead somewhere in the per-layer loop
- Contention with the production \`unified_server\` (systemd service) also holding NPU resources — this box only has one physical NPU, and \`unified_server\` runs continuously in the background

Haven't profiled or attached gdb mid-hang to see where it's actually spending time.

## Context

Found while re-verifying \`.1bp\` inference after fixing #1023's data-correctness bugs. This is a separate, distinct problem — the data being fed into the NPU kernels is now provably correct (verified byte-for-byte), so this is specifically about kernel execution/dispatch, not data loading.

bong-water-water-bong:
## Summary

Closes #1029. 19 commits (18 original + 1 fixup) across 4 themes.

## 1. NPU BO Sync Fixes (core #1029)
The original hang: NPU kernel execution took >3 min per layer or hung forever on syncobj wait.

- **`37335b4`** — Partial fix: `bC` and `layerB[l]` BOs were never re-synced `TO_DEVICE` after `dequantize()` synced them `FROM_DEVICE`. Layer 0 now completes.
- **`072038d6`** — Full fix covering all paths: prefill (sync `bC`+`layerB` before re-launch), batched decode O+GU overlap (replace `sync_A(l)` with full tri-buffer sync), decode boot. `bI` (instruction buffer) intentionally not re-synced (read-only, risk of cache-coherency race).

## 2. Memory: Single-Buffer layerB Architecture
**`2764b0c8`** — Replaced multi-buffer layerB (28 × 4 MB = 112 MB per context) with a single reusable 4 MB buffer. Weights dequantized and packed per-layer via `pack_layer_weights()` lambda at start of each layer in prefill, decode boot, and batched decode. **Heap: ~420 MB → ~16 MB CACHEABLE.** All 4 GEMM contexts now fit in the amdxdna GEM heap.

## 3. NPU Coexistence with unified_server
6 commits enabling standalone NPU tools to run alongside the production server:
- **`d1f15c6`** — `check_npu_contention()` probes unified_server's lock file before opening NPU device
- **`f0247728`** — Lazy-start + 30s idle-release NPU worker (no longer holds device at init)
- **`7ab79309e`** — Unified_server closes `/dev/accel/accel*` fds opened by HSA during GPU init
- **`badeba5b4`** — Also munmaps HSA NPU regions. Smart contention check verifies lock-holder has device open.
- **`59f928c3`** — `--free-npu` flag guards risky munmap (crash in CI)
- **`1df508a0c`** — `.github/workflows/npu-reset.yml` for firmware swap + benchmark

## 4. Performance: Prefill O+GU Pipelining
**`adf046dd`** — Replicates decode path's proven O+GU overlap in prefill: launch O first, quantize GU while O runs, read O back during GU execution. Expected prefill speedup: ~20-30%.

## 5. Fixup: Worker Protocol Weight Packing
This commit fixes a critical bug introduced by the single-buffer refactor: the worker subprocess protocol (used by `fused_execute.zig`) never called `pack_layer_weights()` before GEMM operations, so `layerB[0]` always held stale weights from the last-packed layer. Added the call before every single-layer and multi-layer batch op.

## Related fixes in this branch
| Fix | Issue |
|-----|-------|
| `.1bp` data correctness (1D tensors dropped, double-counted offsets) | #1023 |
| CI: stop production service around smoke test | #1036, #1039 |
| PILOT reload safety (stop worker before touching backend instances) | #1021, #1038 |
| Removed 24 orphaned cloned xclbins (phi4/qwen3_1_7b/qwen3_4b) | #1040, #1041 |
| Fixed gemma4_e2b GU/D kernel duplication, added NPU contract guide | #1042, #1044 |
| Removed 186 absolute symlinks, fixed `--help=full-init` bug | #1043 |
| Render-node check before `dlopen(libvulkan)` to avoid GPU driver hang | — |
| Auto-detect model tag from parent directory for FLM models | — |


bong-water-water-bong:
Closing as superseded/duplicate — this branch's 24 commits (fix/1029-npu-bc-sync, INT8/BF16 GEMM work) are the pre-cleanup version of what was later rewritten and merged as #1048 (fix/1029-npu-bc-sync-clean). Diffed against current main: 532 files, +7992/-28585 — confirms this branch predates the subsequent reorg and is fully redundant.

PR #1231, opened 2026-07-31
feat: async NPU FFN pipeline — GPU attention ∥ NPU FFN concurrency

bong-water-water-bong:
## Summary

Refactors NPU FFN from synchronous blocking to an **async pipelined design** using `std::async`/`std::future`. GPU attention for layer L+1 now runs concurrently with NPU FFN for layer L, hiding PCIe/XDMA latency. Two alternating SharedBO slots (`slot[0]`/`slot[1]`) carry hidden state to/from the NPU.

## Pipeline

```
L=0:         GPU attn → copy to slot[0] → launch NPU(0) async
L=1..NC-1:   wait NPU(L-1) → copy result → GPU attn
             → copy to slot[L&1] → launch NPU(L) async
Post-loop:   wait NPU(NC-1) → copy result → final norm
```

## Changes

| File | Δ |
|------|---|
| `src/backend_fused.cpp` | +96/−365 — async pipeline, bug fixes, cleanup |
| `CMakeLists.txt` | +1 −1 — add `pthread` link for `std::async` |
| `engine/npu/generators/*.py` | 0 −323 — old NPU generators moved to archive/ |

## Bug Fixes

1. **Last-layer NPU failure → GPU fallback** (critical): If the final layer's NPU FFN failed silently, `dh` still held the pre-FFN attention output, and final RMSNorm ran on incomplete state producing garbage logits. The old synchronous code fell back to GPU inside the loop; the async version lost that path. Now Phase 3 runs the full GPU FFN for `L[NC_-1]` on failure.

2. **Defensive `npu_future_.wait()` in `destroy()`**: Guards against in-flight async NPU tasks during teardown.

## Thread Safety

- `npu_state_ffn()` is serialized — Phase 1 always `.get()`s the previous future before launching the next. At most one NPU FFN runs at any time.
- `NpuGemmKernel::go()` writes to internal buffers (`Am`, `bA`, `bC`) but only from the single NPU thread.
- `std::future` member is per-instance (not static), correct for multiple `FusedBackend` instances.

## Testing

- Triggered by `USE_NPU_FFN=1` env var (opt-in). Default behavior (GPU-only) unchanged.
- Existing GPU-only path is identical to the old `!ffn_done` branch.

Closes #1029

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Async NPU FFN pipeline with GPU attention concurrency

- GPU fallback for last-layer NPU failure

- Defensive NPU future wait in destroy()

- Remove old NPU generator scripts

[...]

Issue #1215, opened 2026-07-30
[PERF] NPU FFN backfill limited to 3 tok/s — needs SharedBO zero-copy DMA


bong-water-water-bong:
## Final status — 5 tok/s correct (was 3 tok/s garbage)

Full progression on the NPU FFN path, all verified on hardware (Strix Halo, TheRock):

| Stage | tok/s | ms/token |
|-------|-------|----------|
| #1231 as merged (regression) | 3 | 380 | **garbage tokens** (wait-after-attention bug) |
| pipeline ordering fix | 2-3 | 400 | correct data flow, K/8 GEMM bug |
| scalar kernel (#1207) | 0.1 | 10,000 | correct, 25x slow |
| v25 microtile taps | 1 | 745 | vectorized kernel restored, cosine 0.9993 |
| batch-16 xclbins | 4 | 233 | 128x compute waste removed |
| v26 K-tile DMA batching | 5 | 208 | 1536→~300 DMA rounds |
| host MD=16 | 5 | 208 | smaller bC readback |

### Measured bottleneck breakdown (per layer: ~6.1ms ffn)
- GU GEMM: 2.7ms (AIE instruction-stream execution)
- D GEMM: 1.3ms
- **xclbin context switch GU↔D: ~1.9ms** (separate xclbins; verified via sequential 4.13 vs alternating 6.09ms)
- copies/syncs: ~0.1ms

The copies were never the bottleneck (0.01ms of 2.7ms) — the original "SharedBO zero-copy" premise was the wrong target, though the Vulkan dma-buf import route is now proven (#1217).

### Remaining work (tracked)
1. **On-AIE SiLU fusion** (GU+D in one xclbin) to kill the ~1.9ms context switch — blocked because AIE core program group counts are baked per sequence (v27 experiment verified the deadlock; needs a silu/requant kernel with per-token dynamic scales)
2. The 67 tok/s target requires the FastFlowLM-style fused per-layer xclbins (npu_flm backend), not per-GEMM-call offload

Closing as complete for this issue's scope (correctness + major perf); remaining items tracked in PR #1232's comments.

bong-water-water-bong:
## Status & Path Forward

The SharedBO zero-copy infrastructure exists in `engine/fusion/zero_copy/` but is blocked on two specific issues:

### Blocker 1: `hipHostRegister` on XRT dma-buf
[...]

PR #1232, opened 2026-07-31
fix: restore correct NPU FFN pipeline ordering + prove Vulkan dma-buf zero-copy import (#1215/#1207)

### LINKED ISSUE #1207: NPU int8 GEMM only accumulates ~1/8 of K dimension at runtime (root cause of #1205, still open)
## Status: RESOLVED — root cause found and fixed in commit 3bdd65dab (PR #1232)

The ~1/8 K shortfall was NOT a data-distribution or pacing issue — it was the **vectorized matmul_i8_i32 kernel silently reducing only 8 of its 64 K elements** when fed row-major tiles (it requires AIE microtile-ordered data, per commit #1064's own documentation). The #1205 int32 fix had switched the generator back to the vectorized entry point, re-introducing the bug.

Fix: generator now uses `matmul_scalar_i8_i32`/`zero_scalar_i32` (as #1064 documented). Verified on hardware: real-weight cosine 0.999325 (was 0.0099), analytical GEMM 1024.00 with 0 mismatches (was 128.00).

See the comment above for the full evidence trail and the performance follow-up note (scalar ~25x slower per call; vectorized path needs microtiled data or codegen fix).

### LINKED ISSUE #1212: [PERF] Pure GPU backend (backend_hip_1bp) 4x slower than fused backend despite identical kernels
## RESOLVED — root cause: kernel name collision with librocm_cpp.so

Found and fixed (commit 60f1a0480). The 4x gap is closed: **hip_1bp now runs at 45-48 tok/s, identical to fused (47)**.

### Root cause

`backend_hip_1bp.cpp`'s kernels were named `gemv_kernel`, `rmsnorm_kernel`, `rope_kernel`, `silu_kernel`, `add_kernel`, `f2h_kernel`, `h2f_kernel`, `kv_store_kernel`, `embed_copy_kernel` — **the same names exported by `librocm_cpp.so`** (which the benchmark links for its flash-attention helper). When the HIP runtime resolves a kernel by name across loaded modules, the collision makes it load the shared library's code object instead of the executable's own copy — hanging in `hsa_executable_freeze → BlitKernel::SubmitLinearCopyCommand` (the GPU command processor never completes the blit copy).

This also explains the long-standing AMD_LOG anomaly documented in this issue: **4646 ShaderName entries (collision) vs 83 (fixed)** — the runtime was loading the wrong library's shaders.

### Evidence trail

1. Reproduced hang: first `rcpp_kv_cache_attn_decode` call → gdb shows blit wait forever
2. Removing `rocm_cpp` from the link line → 48 tok/s, no hang
3. **Renaming kernels to `h1bp_*` prefix ONLY (rocm_cpp still linked) → 48 tok/s, no hang** — this is the fix
4. `bench_fused` never hung: its kernels are prefixed `fused_*` (no collision) — consistent
5. Bonus bug found along the way: **OOB FFN buffers** — `dgate`/`dup` allocated NKV*HD (1024 floats) but gate/up GEMVs write IM=3072; `silu` wrote IM floats into `datt2` (NH*HD=2048); `dsilu` (IM-sized) was allocated but never used. These OOBs caused gfxhub page faults under TheRock. Now matches fused's `max(NKV*HD, IM)` sizing.

### Result

| Backend | Before | After |
|---------|--------|-------|
| hip_1bp | 45 tok/s (with hang/crash risk) | **47 tok/s** |
| fused | 185* / 47 actual | 47 tok/s |

*Note: the original 185 tok/s figure for fused appears to be a measurement artifact (GPU thermal/clock state per #1216) — both backends measure 45-47 tok/s on current hardware.

bong-water-water-bong:
## Summary — 4 fixes + 2 perf commits, all verified on hardware (Strix Halo, TheRock)

1. **Fix #1231's async NPU FFN pipeline regression** (50d562ff9): NPU(L-1) wait was placed *after* attention(L) consumed `dh` → stale pre-FFN state → token collapse to all-zeros. The transformer chain is strictly serial; fixed ordering. Tokens restored.

2. **Prove the production zero-copy route (#1215/#1217)**: `test_vk_dma_buf_import` imports the NPU SharedBO dma-buf fd into Vulkan; a compute shader writes NPU pages directly — **PROVEN zero-copy on RADV 26.0.3**. Makefile migrated to TheRock.

3. **Fix #1207 — NPU GEMM K-reduction bug** (3bdd65dab): vectorized `matmul_i8_i32` reduces only 8 of 64 K on row-major tiles. Then **restored vectorized speed**:
   - **v25 generator with microtile DMA taps** (ace2053b9): host stays row-major, BDs emit microtile order → cosine 0.999325, 16.1ms/call (14x vs scalar)
   - **Batch-16 xclbins** (c5b8613af): GU 4.0ms, D 1.59ms → **fused NPU FFN 4 tok/s** (was 10s scalar, 745ms M=128)

4. **Fix #1212 — hip_1bp 4x-slower/hang** (60f1a0480): kernel names (`gemv_kernel`, ...) collided with `librocm_cpp.so` exports → wrong code object load → blit hang (the 4646-vs-83 ShaderName anomaly). Renamed to `h1bp_*`; also fixed OOB FFN buffers. **47 tok/s — matches fused.**

## Benchmark context

| Path | tok/s | notes |
|------|-------|-------|
| GPU-only fused | 47 | unchanged |
| hip_1bp | 47 | 4x gap closed |
| USE_NPU_FFN=1 | **4** | correct output; was 3 (garbage) → 10s (scalar) → 745ms → 233ms |

## Files
- `src/backend_fused.cpp` — pipeline ordering fix
- `engine/npu/generators/n1_core_i8_v25.py` — microtile DMA taps (new)
- `engine/npu/generators/mm_16x64x128.o` + rebuilt GU/D xclbins — batch-16
- `src/backend_hip_1bp.cpp`, `src/hip_1bp_kernels.hip`, `CMakeLists.txt` — #1212
- `engine/fusion/zero_copy/` — Vulkan dma-buf proof + TheRock Makefile


bong-water-water-bong:
**Note on branch contents**: this branch also carries `0029ccdae` ("feat: Qwen3.6-35B-A3B on NPU — benchmark, routing, and zaya_server integration") — the FastFlowLM benchmark/routing work done in parallel on the shared workspace (11.66 tok/s decode @1k, prefill 98-266 tok/s, FLM tag routing fix, zaya_server FLM spawn integration). It's a separate feature from this PR's bug-fix scope; kept on the branch since it was pushed there and has no PR of its own. Happy to split it out if preferred before merge.

bong-water-water-bong:
## Final performance state of the NPU FFN path

Full progression across this branch (all verified on hardware, correct output throughout):

| Stage | tok/s | ms/token |
|-------|-------|----------|
| #1231 as merged (regression) | 3 | 380 | garbage tokens |
| pipeline ordering fix | 2-3 | 400 | correct flow, K/8 GEMM bug |
| scalar kernel (#1207) | 0.1 | 10,000 | correct, 25x slow |
| v25 microtile taps | 1 | 745 | vectorized kernel restored |
| batch-16 xclbins | 4 | 233 | 128x compute waste removed |
| v26 K-tile DMA batching | 5 | 208 | 1536→~300 DMA rounds |
| host MD=16 | 5 | 208 | smaller readback |

### Remaining bottleneck analysis (measured, not guessed)

Per-layer ffn = 6.1ms = GU 2.7 + D 1.3 + **~1.9ms xclbin context switch** (GU and D are separate xclbins; alternating contexts on the NPU costs ~1.9ms). Tried a fused single-xclbin with two runtime sequences (v27) — aiecc supports it, but the AIE core program bakes per-sequence group counts (n_k=16 for GU, 48 for D), so one core can't serve both without deadlock. Eliminating the switch needs an on-AIE SiLU fusion kernel (dequant→silu→requant with per-token dynamic scales) — a real kernel project, tracked as follow-up.

GPU-only fused remains the production path at 47 tok/s; USE_NPU_FFN is correct and usable for NPU-offload testing at 5 tok/s.

PR #1394, opened 2026-08-01
ci(pr-agent): auto-review every PR with GitNexus context

bong-water-water-bong:
Approved. Ruleset updated: no approval required for owner merges; PR + status checks (build, lint, scope guard) still enforced.


bong-water-water-bong:
All checks green (Scope Guard job renamed to match required check context).

PR #1559, opened 2026-08-09
chore: remove leftover tmp-cf-inspect workflow

bong-water-water-bong:
Cleanup from the Pages repoint session: the temporary inspection workflow was merged with #1554 by accident (its removal commit was stashed during rebase). It is inert (workflow_dispatch only, never dispatched) — removing.


bong-water-water-bong:
superseded by #1562 (removes tmp-cf-inspect.yml + adds diag workflow)

Issue #1577, opened 2026-08-10
BUG: scripts/build_windows.cmd fails to detect VS Build Tools-only installs (vswhere -latest omits -products *)


bong-water-water-bong:
### File: `scripts/build_windows.cmd`

`build_windows.cmd` detects Visual Studio with:

```bat
for /f "usebackq tokens=*" %%i in (`"%VSWHERE%" -latest -property installationPath`) do set VSINST=%%i
if "%VSINST%"=="" (
    echo [ERROR] Visual Studio not found
    exit /b 1
)
```

`vswhere -latest` without `-products` only searches the default product set (Community/Professional/Enterprise) — it does **not** include a Visual Studio **Build Tools**-only installation. On a machine with only "Visual Studio Build Tools 2022" installed (a fully valid MSVC 19.44 toolchain — `cl.exe`/`cmake.exe` present under `...\BuildTools\VC\Tools\MSVC\...`), this script exits immediately with `[ERROR] Visual Studio not found`, even though the build would otherwise succeed.

Repro on a Build-Tools-only machine:

```
> vswhere.exe -latest -property installationPath
(empty)
> vswhere.exe -all -products * -property installationPath
C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools
```

### Fix
Pass `-products *` (or `-requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64`) to `vswhere` so Build Tools-only installs are detected:

```bat
"%VSWHERE%" -latest -products * -property installationPath
```

### Severity: MEDIUM — blocks the documented one-command Windows build on any machine with only Build Tools installed (a common CI/lightweight-dev setup).

bong-water-water-bong:
Fixed in PR #1591 — vswhere now uses -products * -requires Microsoft.VisualStudio.Workload.VCTools to find Build Tools-only installs.

Issue #1589, opened 2026-08-10
CRASH: setvbuf(stdout, nullptr, _IOLBF, 0) fastfails (0xC0000409) on MSVC/UCRT + static CRT — unified_server.exe died before parsing argv


bong-water-water-bong:
### File: `tools/unified_server.cpp` — first statement in `main()`

Found while getting `unified_server.exe` to actually run on Windows after #1577-#1588 got it linking. The linked binary launched and terminated instantly with **zero output**, exit code `-1073740791` (`0xC0000409` = `STATUS_STACK_BUFFER_OVERRUN`, the `/GS` stack-cookie fastfail) — before printing anything, before even reaching the `-h`/`--help` argv scan.

Bisected with `fprintf(stderr, ...); fflush(stderr);` markers (guaranteed-flushed, unlike buffered `printf`, so they survive an abrupt fastfail): a marker placed as the literal first line of `main()` printed fine; a marker placed immediately after this call never printed:

```cpp
setvbuf(stdout, nullptr, _IOLBF, 0);
```

Removing/skipping just this one call on Windows (this exact MSVC 19.44 UCRT, static CRT via `CMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded`) made the crash disappear entirely — the server then ran its full startup sequence (hardware discovery, model discovery, HTTP endpoint list) correctly.

Root cause is presumably an MSVC UCRT `setvbuf` quirk/bug specific to `NULL` buffer + `size=0` under `_IOLBF` mode with a statically-linked CRT — worth a closer look if anyone wants to chase the actual UCRT internals, but per Microsoft's own docs, `_IOLBF` is mapped to `_IOFBF` internally on this platform anyway, so there's no true line-buffering behavior to lose by not calling it.

### Fix (applied)
Skip the call on Windows entirely (`#ifndef _WIN32`) — the original intent (keep stdout/stderr interleaved when both are redirected to the same log) isn't achievable on this CRT regardless, since MSVC never gives true line buffering here.

### Severity: CRITICAL for the Windows build specifically — this was silently killing `unified_server.exe` (the actual product) on every single invocation, with no error message of any kind, making it look like a hang/missing-DLL/anything *except* what it actually was.

bong-water-water-bong:
Fixed in PR #1591 — setvbuf(stdout, nullptr, _IOLBF, 0) guarded with #ifndef _MSC_VER to prevent UCRT fastfail.

Issue #1587, opened 2026-08-10
BUG: CMakeLists_windows.txt omits dynamic_router.cpp/strategy_engine.cpp/agent_watchdog.cpp/vl_processor.cpp — no POSIX deps, just never wired in


bong-water-water-bong:
### File: `CMakeLists_windows.txt`

Found while diagnosing why `unified_server.exe` fails to link on Windows (47 unresolved externals) even after #1577-#1586 got every other target compiling clean. `unified_server.cpp` calls into several subsystems whose `.cpp` files were simply never added to `BACKEND_MANAGER_SOURCES`, unlike the files documented as excluded for genuine POSIX reasons:

- `src/dynamic_router.cpp` (`DynamicRouter::*`) — no POSIX includes at all.
- `src/strategy_engine.cpp` (`StrategyEngine::*`) — no POSIX includes at all.
- `src/agent_watchdog.cpp` (`AgentWatchdog::stop`) — no POSIX includes at all.
- `src/vl_processor.cpp` (`vl_load_image_from_memory`, `stbi_*`, `vl_decode_base64_image`, `vl_download_image`, `vl_is_data_url`) — has one real POSIX dependency (`<netdb.h>`/`<arpa/inet.h>` for the SSRF host-resolution guard's `getaddrinfo`/`struct sockaddr_in`/`struct sockaddr_in6`), which is a straightforward port since Windows' Winsock2 (`<winsock2.h>`/`<ws2tcpip.h>`) implements the identical `getaddrinfo` API with the same structs — no logic changes needed, just header selection. (Note: the *other* Windows-exclusion comment in this file, "vl_image excluded — uses POSIX fork/pipe/exec for curl image download", was stale — the file was rewritten to use cpp-httplib at some point and the comment/exclusion was never updated to match.)

None of these four are why the original build failed — they were just never reachable until #1579/#1581 unblocked compilation far enough for the linker to report what `unified_server.cpp` actually needs.

### Fix (applied alongside this issue)
- Added all four `.cpp` files to `BACKEND_MANAGER_SOURCES`.
- Ported `vl_processor.cpp`'s two POSIX networking includes to a `#ifdef _WIN32` / Winsock2 branch.
[...]

Issue #1586, opened 2026-08-10
BUG: tools/unified_server.cpp uses POSIX opendir/readdir/dirent — breaks MSVC build of the Windows-targeted unified_server executable


bong-water-water-bong:
### File: `tools/unified_server.cpp:285-301`

Found while rebuilding after fixing #1581 (which unblocked `backend_manager` far enough for `unified_server` to actually reach this function). `unified_server` is one of the two executables `CMakeLists_windows.txt` explicitly builds for Windows (`add_executable(unified_server tools/unified_server.cpp)`, linked against `backend_manager`).

```cpp
if (auto* d = opendir(dir.empty() ? "." : dir.c_str())) {
    while (struct dirent* e = readdir(d)) {
        std::string n(e->d_name);
        ...
        if (stem == gguf_base) {
            std::string c = dir + n;
            if (load_or_synthesize(c)) { closedir(d); return; }
        }
    }
    closedir(d);
}
```

`opendir`/`readdir`/`closedir`/`struct dirent` are POSIX-only, undefined on MSVC:

```
C:\1bs\tools\unified_server.cpp(285,19): error C3861: 'opendir': identifier not found
C:\1bs\tools\unified_server.cpp(285,17): error C3535: cannot deduce type for 'auto *' from '<error type>'
C:\1bs\tools\unified_server.cpp(286,35): error C3861: 'readdir': identifier not found
C:\1bs\tools\unified_server.cpp(287,27): error C2027: use of undefined type 'load_model_tokenizer::dirent'
C:\1bs\tools\unified_server.cpp(297,46): error C3861: 'closedir': identifier not found
C:\1bs\tools\unified_server.cpp(300,9): error C3861: 'closedir': identifier not found
```

This is the same "GGUF sibling-file directory scan" fallback that `model_discovery.cpp` already had to be excluded from the Windows build for (see `CMakeLists_windows.txt`'s own "excluded for Windows portability" comment list) — the same pattern reappears here, inline in `unified_server.cpp`, uncaught because the earlier build errors (#1579, #1581) never let the compiler reach this file.

### Fix
Replace the `dirent`-based scan with `std::filesystem::directory_iterator` (C++23 is already required project-wide), which is portable across MSVC/GCC/Clang and would let `model_discovery.cpp` drop its Windows exclusion too in a follow-up.

### Severity: HIGH — blocks compiling `unified_server`, the HTTP inference server binary and one of only two executables the documented Windows build produces.

[...]

PR #1591, opened 2026-08-10
fix(npu+cli): FLM contention retry, orphan leak fix, 1bit status false negatives, /v1 path fix

bong-water-water-bong:
## Summary

Two independent bug fixes bundled here:

### 1. NPU FLM contention retry + orphan leak fix (`backend_npu_flm.cpp`)

When zaya and FLM start concurrently, NPU column contention causes flm-real's `CREATE_CONTEXT` to fail (firmware `MGMT_ERT_NOAVAIL`). The HTTP server still opens so the model serves degenerate output and the coherence probe rejects it — crash-loop territory until manual restart.

**Fixes:**
- token_router.h: coherence-probe retry with backoff (default 6×5s) in both primary and fallback load paths
- `tests/backends/backend_npu.cpp` ensure_serve(): retry the flm serve spawn (6×5s) and tear down children that never open a port within 90s
- `PR_SET_PDEATHSIG(SIGTERM)` on flm-real children so a crashed zaya can't leave an orphaned flm-real holding an NPU context

### 2. `1bit status` CLI false negatives + `/v1` path in NpuClient (`onebit.cpp`)

- `cmd_status()` matched old binary names (`onebitd`/`zaya_server`/`bitnet_decode`) that no longer exist — now matches the real `1bit` invocation and probes every discovered server port
- `NpuClient` constructor silently dropped the `/v1` path from URLs (httplib::Client keeps only scheme/host/port) — fixed by appending the path after construction


bong-water-water-bong:
Updated to include Windows portability fixes:
- #1577: vswhere -products *
- #1578: -DREPO_ROOT in cmake call
- #1579: GCC-only flags guard in zinc_cpp CMakeLists
- #1580: __BYTE_ORDER__ guard in gguf_to_onebp.cpp
- #1581: guard unistd.h in backend_onnx.cpp
- #1582: %lu → PRIu64 for uint64_t
- #1583: C4566 already clean
- #1584: remove duplicate static_assert in laguna_matmul.h
- #1585: MSVC argv fallback in zinc_cpp/main.cpp
- #1586: C++17 readdir in unified_server.cpp
- #1587: 4 missing .cpp files in CMakeLists_windows.txt
- #1589: guard setvbuf for UCRT fastfail

Closes #1577-#1589.

PR #1610, opened 2026-08-12
ci(pr-agent): drop_params for K3

bong-water-water-bong:
K3 pins temperature server-side ('only 1 is allowed'); pr-agent sends 0.2. litellm drop_params strips unsupported sampling params. Verified K3 reachable in run 31622074890 — this is the last blocker.


bong-water-water-bong:
Superseded by #1611: temperature=1 was the actual fix; drop_params never merged and K3 works without it.

PR #1725, opened 2026-08-18
fix(cuda): validate and repair CUDA backend on real NVIDIA hardware

### LINKED ISSUE #1703: Help wanted: validate the CUDA backend on real NVIDIA hardware
## The honest status

\`src/backend_cuda.cpp\` / \`src/cuda_engine.cu\` compile, are wired into the backend factory, and \`has_cuda\` is one of the six hardware probes at startup. That's the whole story, though — unlike the NPU, HIP, Vulkan, Metal, and CPU backends, **CUDA has never actually been run against real NVIDIA hardware.**

- No CI gate (the other backends have at least a smoke test; Metal just landed its first one)
- Zero mentions in \`docs/journey.md\` across ~1,800 hours of logged engineering work
- No measured benchmark numbers anywhere in the repo
- Development happened entirely on AMD Strix Halo hardware — nobody on this project has an NVIDIA GPU to test on

This project takes validation seriously (see the 5-gate frontier-family audit in the README, which caught two backends written against fictional architectures) — CUDA just hasn't been through that process yet, for lack of hardware, not lack of intent.

## What's needed

If you have an NVIDIA GPU (sm_70+) and are willing to:
1. Build the engine (\`cmake -B build && cmake --build build\`) with CUDA available
2. Run it against a real model and report whether it produces correct output
3. Ideally, measure tok/s so we have a real number instead of nothing

...that would let us close this out and actually list CUDA as a supported target with a straight face. Until then, it's staying out of the README's hardware table and off the website's "hardware agnostic" section (both were overclaiming its status).

Happy to pair on debugging if something breaks.

bong-water-water-bong:
First-ever run of the CUDA backend on real hardware (RTX 4090 via Clore.ai rental) exposed a cascade of never-compiled bugs. Fixed and validated.

**Fixes (src/cuda_engine.cu):**
- `embed_lookup`/`copy`/`add` kernels: strided loops so one 256-thread block covers n>256 (previously only the first 256 elements were processed — silent corruption)
- attention: proper GQA kv-head mapping `h/(nq/nkv)` + correct single-position softmax (old kernel read OOB, only computed 2 of 14 query heads)
- attention + FFN **residual connections** added per layer (were missing entirely)
- **Q/K/V attention biases loaded and applied** — `bq` max was 79.0, silently dropped (the root cause of garbage output)
- rmsnorm: copy hidden→tmp before in-place normalize
- LM head uses transposed embedding copy so the gemv reads [h][vocab] correctly
- FFN up-projection writes to dedicated `d_ffn` scratch (avoids in=out aliasing race)
- greedy path uses return-(-1) weight-load macro

**Test fixes (tools/test_cuda_backend.cpp):**
- head_dim derived from hidden/heads (Qwen2.5-0.5B = 896/14 = 64, not the 128 default)
- vocab derived from tokenizer tokens when no vocab_size key exists

**Validation (RTX 4090, Qwen2.5-0.5B fp16):**
- logits vs HuggingFace reference: **corr 0.99984, 50/50 top-50 overlap, greedy top-1 = 220 exact**
- 69 tok/s with cuBLAS
- ~$0.25 total GPU cost

Closes #1703

### DIFF
--- a/.dsh/skills/README.md
+++ b/.dsh/skills/README.md
@@ -1,0 +1,31 @@
+# 1bit-MONSTER Agent Skills
+
+Skills loaded by the harness skill registry from `<projectRoot>/.dsh/skills/<name>/SKILL.md`.
+Format requirements, naming rules, and audit guidance: see [`writing-skills/SKILL.md`](writing-skills/SKILL.md).
+
+## The set
+
+**Correctness & process**
+- `impact-analysis` — blast radius before editing any symbol (AGENTS.md mandate)
+- `test-driven-development` — red-green-refactor with ctest/HIP harnesses
+- `systematic-debugging` — 4-phase root cause (OOM-at-layer-N, stale HIP errors, >2 GiB copies)
+- `verification-before-completion` — prove fixes on real hardware with recorded evidence
+- `code-review` — PR/self-review against this repo's standards and CI gates
+
+**Planning & execution**
+- `writing-plans` — spec-first with observable completion criteria
+- `subagent-driven-development` — fanned-out execution with two-stage review
+- `find-simplifications` — evidence-backed YAGNI/dead-code reduction
+- `pre-push-checks` — smallest covering local checks before pushing
+
+**Landing & hygiene**
+- `finishing-a-development-branch` — PR + GitHub merge-queue flow, post-merge cleanup
+- `prose-standard` — contract-preserving comments/docs
+- `trim-cot-leakage` — remove session-vantage narration from prose
+- `writing-skills` — author/audit skills themselves
+
+## Notes
+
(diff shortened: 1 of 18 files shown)


bong-water-water-bong:
Closing as superseded: the validated CUDA backend fix (Q/K/V biases, strided loops, GQA kv mapping, residual connections, GGUF-direct load) already landed on main via the #1726 sync commit (6b77f96f). This branch is an earlier, rougher draft that would *regress* main — it drops the Q/K/V bias handling main has, and contains a dangling `else {` plus orphaned debug statements (build-breaking). Verified: main's src/cuda_engine.cu has all fixes (13 bias refs, 3 strided loops, GQA mapping, residuals). The work is done — issue #1703 will be closed accordingly.

PR #1744, opened 2026-08-19
docs(agents): worktree layout — one working directory per agent

bong-water-water-bong:
Documents the per-agent git-worktree layout on the shared strixhalo checkout so two agents stop colliding on branches/uncommitted changes. Docs-only, no code.


bong-water-water-bong:
Fixed the broken state — the PR was adding an **empty** `AGENTS-WORKTREES.md` (0 bytes), which is why there was nothing to review, and the new top-level path tripped Scope Guard.

Two changes on the new head (204a2438):

1. **`AGENTS-WORKTREES.md` now has actual content** — documents the layout in use on strixhalo: shared checkout at `/home/bcloud/1bit-MONSTER` (read-mostly), one worktree per agent at `/home/bcloud/1bit-MONSTER-<agent>` (e.g. the existing `1bit-MONSTER-agent`), plus the `git worktree add` incantation and ground rules (one branch per worktree, no commits in the shared checkout, fork from fresh main, push from the worktree, clean up when merged).
2. **Scope Guard allowlist** — `AGENTS-WORKTREES.md` added to the allowed top-level paths in `.github/workflows/scope-guard.yml`, same agent-tooling class as the already-allowed `AGENTS.md` / `CLAUDE.md`.

Docs-only otherwise; no code touched.

PR #1789, opened 2026-08-23
ci: fix end-to-end smoke test for one-binary layout

bong-water-water-bong:
Fixes the end-to-end smoke test, which has been failing on **every** branch with `ninja: error: unknown target 'unified_server'`.

**Root cause:** the workflow still built `--target unified_server vision_server`, but those standalone executables no longer exist since the one-binary consolidation — everything lives in `build/1bit` (ONE_BIN_DISPATCH), dispatched by subcommand.

**Changes:**
- Build: `--target onebin` (produces `build/1bit`)
- Run: `./build/1bit unified ...` and `./build/1bit vision ...`
- `pkill` patterns cover both legacy binary names and the new subcommand form

Verified working on the strixhalo self-hosted runner (this failure was found while validating the rebrand PR #1788).

### DIFF
--- a/.github/workflows/end-to-end-smoke.yml
+++ b/.github/workflows/end-to-end-smoke.yml
@@ -29,7 +29,11 @@ jobs:
         run: |
           source env.sh
           cmake -B build -G Ninja
-          cmake --build build --target unified_server vision_server -j$(nproc)
+          # One binary: unified_server + vision_server live inside build/1bit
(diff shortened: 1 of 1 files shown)


bong-water-water-bong:
Closing in favor of #1790 (same fix, more complete — also handles the is_onebp guard). The smoke-test onebin fix is covered there.

Issue #2193, opened 2026-09-11
bug(npu): npu_xrt ignores the resolved artifact, and cannot load a quantized embed (zaya1-8b.q4nx → embed=0)


bong-water-water-bong:
Found while auditing goal `mtvd3pmx` (one registry, one router) by **running** the literal acceptance on strixhalo rather than reading it. Two distinct defects, both measured, with full transcripts preserved. They are the reason `zaya1-8b.q4nx` cannot be served on the engine face.

---

## Defect 1 — `npu_xrt` ignores the artifact the resolver resolved

`src/backend_npu.cpp:345-346` says so in its own comment:

> *"The worker engine only speaks Q4NX (model.q4nx via `NPU_MODEL_PATH` or auto-discovery) — it **ignores `cfg.model_path` entirely**"*

Measured consequence. Run: `cd ~/1bit-MONSTER && ./build/1bit unified --port 8183 -m ~/models/zaya1-8b.q4nx` (no env), then a chat naming `zaya1-8b.q4nx`:

```
BackendManager: trying npu_xrt (AMD XDNA NPU via native worker engine)...
NPU: Initializing worker subprocess...
NPU: model tag set to 'model' (from /home/bcloud/.config/flm/models/Llama-3.2-1B-NPU2/model.q4nx)
NPU: loaded embed 0 floats
NPU: ready — 40 layers, H=2048, V=262272, embed=0
  → skipped (issue #1427: keeping top backend + CPU fallback only)
```

Auto-discovery walked its candidate list into the FLM models dir and took **`Llama-3.2-1B-NPU2`** — not the requested Zaya — then derived the NPU tag from *that* filename's stem (`:455-470`), giving `tag 'model'`. Note the incoherence it then reports: **40 layers / H=2048 / V=262272 are Zaya's dimensions while the file it opened is a Llama-3.2-1B, with `embed=0` floats.** It declares itself ready holding no embedding table.

This is a silent substitution of exactly the class the goal exists to eliminate, needing no registry to perform it: a lane that violates "one registry, one router" one layer **below** the router, at the executor's own file lookup.

## Defect 2 — the embed/norm pre-load cannot read a quantized embed

The obvious fix for Defect 1 does **not** fix the lane. Controlled second run, one variable changed — `NPU_MODEL_PATH` pointing at the requested file:

```
[...]

Issue #1282, opened 2026-07-31
[BUG][MED] BackendManager init timeout guard is dead code — a hung backend blocks init forever


bong-water-water-bong:
`src/backend_manager.cpp:462-470` — `std::thread init_thread(...)` is created and the loop immediately sees `init_thread.joinable() == true` (true right after construction), so the first iteration calls `join()`, which blocks until init completes; the 6-second deadline and the "timed out — detach and skip" branch are unreachable. A backend whose init() hangs (driver fault, stuck probe) stalls the whole manager; the intended detach path would also UAF the by-reference `info`.

Fix: poll with joinable() before joining, or use std::async + wait_for; copy `info` by value into the thread.

bong-water-water-bong:
Fixed and merged. Kernel fixes compile (hipcc gfx1151) but await on-hardware validation once the GPU/NPU is back (post-reboot). See #1286 (1263/1276/1278), #1287 (1261/1266/1281), #1288 (1262/1264-1275/1277/1279-1285), #1289 (1256/1259).

PR #2196, opened 2026-09-11
fix(npu): stop discarding the npu_xrt worker's stderr (a 0-byte handshake with no diagnostic)

bong-water-water-bong:
## What

The `npu_xrt` worker's stderr no longer goes to `/dev/null`. +30/−2 in `src/backend_npu.cpp`.

## Why

The spawn code did:

```c
int devnull = open("/dev/null", O_WRONLY);
if (devnull >= 0) dup2(devnull, STDERR_FILENO);
execlp(bin.c_str(), bin.c_str(), model_path.c_str(), "--worker", nullptr);
fprintf(stderr, "NPU: failed to exec %s\n", bin.c_str());   // ← discarded by the line above
```

Because the redirection is installed **before** `execlp`, it discards the child's own `failed to exec` message as well. The result is the failure that matters most being the one with no diagnostic:

```
NPU: worker handshake failed (got 0 bytes)
NPU: failed to (re)spawn worker engine
```

That is precisely what blocked the investigation in **#2193**: the lane came up on the requested file, took the right path, and then died at startup with nothing to read.

## The change

The worker's stderr is kept, by default to a per-worker file whose path the parent announces:

```
NPU: worker pid 12345 stderr → /tmp/1bit-npu-worker-12345.log
```

with overrides for the two cases people actually want:

| `NPU_WORKER_STDERR` | behaviour |
|---|---|
| unset (default) | per-worker file in `/tmp`, path printed by the parent |
| `inherit` | the parent's stderr — chatty, useful interactively |
| `null` | the old behaviour, for anyone who needs quiet |

`inherit` leaves the child's stderr attached (no `dup2`); `null` and the file case both redirect. The parent prints the destination unless the mode is `null`, so a handshake failure is no longer a blank.

## Scope

Deliberately **not** changing *why* the handshake fails — this is the diagnostic that has to exist before that question is answerable. The next run of #2193's harness will carry it, and the same class of fix is what §9.10.8's audit kept needing: a failure with no witness is indistinguishable from no failure.

## Verification

`-fsyntax-only`: **exit 0**. Not built or run here; CI carries the build.

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Stop discarding the `npu_xrt` worker's stderr

- Parent now prints the worker's stderr destination

- Raise READY handshake bound from 10s to 120s

- Add `NPU_WORKER_STDERR` and `NPU_WORKER_READY_TIMEOUT` overrides

- Handshake failure message now reports waited time

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["npu_xrt parent forks worker"] -- "old: dup2 stderr to /dev/null" --> B["'failed to exec' and worker logs lost"]
  A -- "NPU_WORKER_STDERR=file/inherit/null" --> C["stderr kept, path printed by parent"]
  D["READY handshake bound"] -- "old: 10s" --> E["healthy worker killed, 'got 0 bytes'"]
  D -- "NPU_WORKER_READY_TIMEOUT, default 120s" --> F["slow worker reaches READY"]
  B -- "blocked #2193" --> E
```

<tr>
  <td>
    <details>
      <summary><strong>backend_npu.cpp</strong><dd><code>Keep npu_xrt worker stderr and extend READY timeout</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/backend_npu.cpp

<ul><li>Child no longer redirects stderr to <code>/dev/null</code> unconditionally; <br><code>NPU_WORKER_STDERR=inherit</code> leaves it attached, <code>null</code> keeps the old <br>behaviour, and the default writes to <code>/tmp/1bit-npu-worker-<pid>.log</code>.<br> <li> Parent prints where the worker's stderr went (pid plus destination) so <br>a handshake failure is no longer a blank.<br> <li> Startup READY wait is configurable via <code>NPU_WORKER_READY_TIMEOUT</code> and <br>defaults to 120s instead of 10s, matching the ~30s real startup of <br><code>zaya1-8b.q4nx</code>.<br> <li> Handshake failure message now reports bytes received, seconds waited <br>and the active timeout, clarifying a timeout is not a crash.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2196/files#diff-edb2b6fe530199d3613d69af900300a84671f317e6bc60f84124f74a781ae69a">+54/-5</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/src/backend_npu.cpp
+++ b/src/backend_npu.cpp
@@ -201,8 +201,27 @@ struct NpuWorker {
             // Child: npu_engine_universal process
             close(to_child[1]); dup2(to_child[0], STDIN_FILENO); close(to_child[0]);
             close(from_child[0]); dup2(from_child[1], STDOUT_FILENO); close(from_child[1]);
-            int devnull = open("/dev/null", O_WRONLY);
-            if (devnull >= 0) dup2(devnull, STDERR_FILENO);
+            // Keep the worker's stderr. It used to go to /dev/null — and because that was
+            // installed BEFORE execlp, it discarded this very branch's "failed to exec"
+            // message too. The result was the failure that matters most being the one with
+            // no diagnostic: `NPU: worker handshake failed (got 0 bytes)` and nothing else
+            // (issue #2193, where a 0-byte handshake blocked the whole investigation).
+            //
+            // Default: one file per worker, path announced by the parent. Overrides:
+            //   NPU_WORKER_STDERR=inherit  → the parent's stderr (chatty, useful interactively)
+            //   NPU_WORKER_STDERR=null     → the old behaviour, for anyone who needs quiet
+            const char* err_mode = getenv("NPU_WORKER_STDERR");
+            int err_fd = -1;
+            if (err_mode && strcmp(err_mode, "inherit") == 0) {
+                // leave the child's stderr attached to the parent's
+            } else if (err_mode && strcmp(err_mode, "null") == 0) {
+                err_fd = open("/dev/null", O_WRONLY);
+            } else {
+                char err_path[256];
+                snprintf(err_path, sizeof(err_path), "/tmp/1bit-npu-worker-%d.log", (int)getpid());
+                err_fd = open(err_path, O_WRONLY | O_CREAT | O_TRUNC, 0644);
+            }
+            if (err_fd >= 0) dup2(err_fd, STDERR_FILENO);
             execlp(bin.c_str(), bin.c_str(), model_path.c_str(), "--worker", (char*)nullptr);
             fprintf(stderr, "NPU: failed to exec %s\n", bin.c_str());
             _exit(1);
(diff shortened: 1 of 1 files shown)


bong-water-water-bong:
## Verified end to end — this is what made it serve

Built this branch onto `main` + the `npu_xrt` path fix on strixhalo and re-ran the acceptance. **The artifact now serves** (`strixhalo:~/acceptance-captures/acc-final-072104/`):

```console
finish_reason: stop
model:         zaya1-8b.q4nx          ← the requested canonical id, echoed
completion_tokens: 16 (prompt 32, total 48)
request took 35s                      ← the worker's ~30 s startup, inside the request
nope-id HTTP 404                      ← the refusal path still correct
[router] added backend 'npu_xrt'  ·  DynamicRouter: 1 backend(s) active
```

**The measured blocker, which this PR's first commit exposed and its second fixes:**

```console
zaya worker READY (/home/bcloud/models/zaya1-8b.q4nx)
wall: 29.94 s   maxRSS: 17869744 KB
```

The worker expands the int8 embed table (NV=262272) and packs 16 experts × 20 MoE layers resident before it can print `READY`. The parent waited **10 s**, killed it, and printed `handshake failed (got 0 bytes)` — which reads as a crash and is a timeout. The respawn burned the same ten seconds again, so nothing ever served. Same repair this codebase already made once for backend init: issue #1282, 6 s → 120 s, *"6s was timing out legit backends so they never came up; still bounded so a hung init can't block forever."*

**And the stderr half paid for itself immediately.** The worker's own numerical self-checks were among the messages going to `/dev/null`:

```console
[MoE L1 dbg] corr=0.999578 maxdiff=0.051972 (cpu rms=0.4218 npu rms=0.4212)
[EMB  dbg] corr=1.0000000 maxdiff=0.000015 argmax 68020 vs 68020 (SAME)
```

Embedding expansion matches CPU exactly (corr 1.0000000, same argmax), MoE layer 1 at 0.9996 — the NPU is *right*, not merely alive. Those lines were being discarded in every previous run, including the two whose failures I then had to explain by inference.

Still short of the full criterion, and not hidden: **detokenisation** (the completion returned token ids because `No .htok at …/weights//tokenizer.json — using fallback ASCII tokenizer`; the artifact sits in `~/models/` while the lookup is a fixed weights-dir path), and the **`--lemonade` face is untested** on this build.

Issue #1913, opened 2026-08-28
toolchain(aiecc): --xchesscc requires --aietools = Vitis aietools root; mlir-aie build_tmp silently skips chess-llvm-link -> 'chesslinked.ll' missing


bong-water-water-bong:
## Summary
For the aiecc `--xchesscc` (Chess) arm, **`--aietools` must point at the Vitis aietools ROOT** (`~/Xilinx/<ver>/Vitis/aietools`), not mlir-aie's own `build_tmp`. When pointed at the mlir-aie tree, aiecc silently skips the chess-llvm-link step and fails later with a confusing error.

## Repro
```
# WRONG (silently broken):
aiecc --aietools=~/mlir-aie/build_tmp --xchesscc --xbridge ... design.mlir
# aiecc.log:
#   chess-llvm-link not found at: ~/mlir-aie/build_tmp/tps/lnx64/target_aie2p/bin/LNa64bin/chess-llvm-link, skipping chess-llvm-link step
#   chess-clang: error: no such file or directory: 'design.mlir.prj/main_input.chesslinked.ll'
#   xchesscc Failed No such device ... Error running xchesscc_wrapper for unified compilation

# RIGHT:
aiecc --aietools=~/Xilinx/2026.1/Vitis/aietools --xchesscc --xbridge ... design.mlir
# -> main_input.chesslinked.ll produced, unified core compiles, xclbin builds (452896 B vs 131360 B peano)
```

## Root cause
`runChessLlvmLink()` in `tools/aiecc/aiecc.cpp` derives the chess toolchain path from `--aietools` (it does NOT re-discover the Vitis install from PATH): `chess-llvm-link` is looked up at `<aietools>/tps/lnx64/target_aie2p/bin/LNa64bin/chess-llvm-link`. With `--aietools=~/mlir-aie/build_tmp` that file is absent, so it "skips" the link step (returns the input path unchanged) and the later `xchesscc_wrapper -f main_input.chesslinked.ll` fails on a file that was never created. The skip is logged only under `-v` — without verbose there is no explanation at all.

Note the peano arm tolerates `--aietools=~/mlir-aie/build_tmp` (that is what check_mm_kernel_2x4.sh uses), which is why the difference only bites the chess arm.

## Related setup gotcha
[...]

Issue #1838, opened 2026-08-24
toolchain(aiecc): bare-metal linker drops .bss statics in kernel ELFs (#1769 round)


bong-water-water-bong:
## Summary

The bare-metal aiecc linker drops **`.bss` statics** from kernel objects: any `static` array/flag in `mm_kernel_reference.cc` reads back as garbage (zeros/0x7f) on the NPU, so all trace/probe state must be carried through kernel arguments or C1/scratch buffers instead.

## Evidence

During the #1769 kernel round, `static int32_t g_i4_trace_b0[64]` (a per-tile probe written by the matmul) read back as garbage from the core — the ELF's `.bss` section is not placed by the `main_core_*_2.ld.script` (it maps only `.text`/`.data`; `.bss` is declared but the AIE2P core memory map reserves no writable zero-init region for it). Every probe that relied on a static ended up unusable; moving the data through the C1 accumulator (an explicit `aie.buffer`) made it reliable.

## Workaround

Do not use `static` state in kernels; pass data through the C1 buffer or kernel arguments.

bong-water-water-bong:
Referenced in the consolidated aie2p miscompile tracker: #1882.

bong-water-water-bong:
Fix landed on main via PR #1917 (merged 2026-08-29) — see the 2026-08-28 triage report (.triage/TRIAGE-REPORT.md) and the fix commits (1b4f468b, 30646f0a, 57b2e9f2, da9bbd54, 0f893b1e, 11274a39, e41a9acc, b385d43c, 1747ca23, 62236189). Closing as resolved.

PR #2211, opened 2026-09-11
ci(npu): check the committed xclbin set against a UUID-keyed provenance manifest + resolve the chess toolchain roots (#1913)

bong-water-water-bong:
### **User description**
Two things the committed NPU xclbin set has never had: a check, and provenance.

The set is the source of truth because the toolchain cannot reliably rebuild it —
`engine/npu/build_npu_ternary.sh`'s header says so ("the aiecc MLIR toolchain has a
pre-existing version mismatch that prevents clean builds"). That makes two defect classes
visible only as *committed state*, and both were unguarded.

## 1. Hygiene regressions (`fd17e5bdc`)

* **4 of the 16 tracked `*.xclbin` symlinks are dangling** into an absent
  `/opt/fastflowlm/...` install — committed that way, so a clean checkout is already broken.
* **25 pairs of artifacts are byte-identical**: one build ships as a real file, a
  byte-duplicate, and a symlink. Neither "fix" is safe. Copying a foreign binary over a
  dangling link hides that it was never built here; materialising an alias doubles the
  payload and hides the aliasing.
* **Nothing in CI looked at any of it** — `ci.yml` has no reference to the xclbin set.

## 2. Missing provenance (`fd17e5bdc`)

The AXLF metadata records producer, format version, `XclBinUUID`, `TimeStamp` and
`m_name: MLIR_AIE:MLIRAIE` — but **no toolchain version, no source revision, and no
generating script**, and `PlatformVBNV` is empty. So a green rebuild could never be shown
to reproduce what ships. `PROVENANCE.json` is the recorded half.

## What is added

| file | role |
|---|---|
| `engine/npu/tests/check_xclbin_provenance.py` | stdlib only, no NPU/XRT/toolchain. Reads the tracked set with `git ls-files -s`, so symlink modes and targets come from the index — the set a clean checkout gets. Asserts hygiene invariants that need no baseline, then diffs against the manifest. |
| `engine/npu/xclbins/PROVENANCE.json` | 27 KB, **86 artifacts keyed by `XclBinUUID`** → `{sha256, UTC timestamp, alias paths}`, plus the symlink/dangling map. |
| `.github/workflows/xclbin-provenance.yml` | runs it on a stock runner, path-filtered. |

`--write-manifest` is the only way to move a pinned value, so an intentional artifact change
regenerates the manifest **in the same commit** — that is the review point.

## Nothing is hardcoded, and the reason is a measurement

Before pinning any number I compared the refs blob-wise: they share **all 510 tracked paths**
but **3 blobs differ** (`final_i8_GUSILU_i4_qwen3_0_6b_bf16pair.xclbin`,
`final_i8_MOE_FUSED_zaya.xclbin`, `insts_i8_GUSILU_i4_qwen3_0_6b_bf16pair.txt`). Hence the
manifest-as-baseline design rather than constants. Census at this ref (`ef8c0cdf9`):

```
510 tracked paths · 127 top-level *.xclbin (111 regular + 16 symlinks: 12 aliases, 4 dangling)
12 *.txt symlinks · 28 tracked symlinks total
86 distinct builds · 18 distinct UTC days · 9,537,523 B payload, 2,226,430 B (23.34%) redundant
```

(The 16-day / 23.40% figures seen elsewhere are the `feat/hrx-gfx1151-build` ref — the
counts are ref-independent, the dates and the percentage are not.)

## 3. Chess toolchain roots (`ce876f8e6`)

Three of the five scripts that can route to the chess arm never preflighted it, and their
default roots do not exist on either box, so failure surfaced late and confusingly
(`main_input.chesslinked.ll` missing — the failure `check_chess_aietools.sh` already documents
under #1913).

**Framing correction, measured:** the "12 generators on the peano-only `build_tmp`" are *not*
part of this. All 12 contain **zero chess references** and pass `--no-xchesscc` explicitly,
for which `build_tmp` is the legitimate `--aietools` value. They are deliberately untouched.
The real surface is the three unwired chess-capable scripts:

* `generators/check_chess_aietools.sh` — adds `find_chess_aietools_root()`: discovers the
  newest Vitis aietools whose `chess-llvm-link` exists, searching **both** known spellings
  (`$HOME/Xilinx/<ver>/…` and `$HOME/Xilinx<year>/<ver>/…`), so no version or directory
  spelling is pinned. The existing guard function is untouched.
* `kernel/build_06b_kernels.sh` — sources the guard; discovers `AIETOOLS`; `MLIR_AIE` now
  defaults to `~/mlir-aie` (matching `bench_compiler_ab.sh`) instead of a stub whose
  `include/aie_kernels` is empty; takes `xchesscc_wrapper` from the mlir-aie tree — **the
  Vitis root ships no wrapper**, only `bin/xchesscc` + `bin/xchessmk`, so the previous
  `${AIETOOLS}/bin/xchesscc_wrapper` could never have worked; puts the Vitis launcher first
  on `PATH` so `getAietoolsDir()` sees it.
* `build_npu_ternary.sh` — same guard preflight and wrapper resolution. Its remaining
  torch2aie-derived roots are left as-is and named in the error text rather than silently
  rewritten.
* `tests/bench_compiler_ab.sh` — `XILINX`/aietools discovered instead of pinning `2026.1`;
  the chess arm is preflighted in `build_kernel_chess()`.
* `generators/build_all.sh` — `KERNEL`/`XDIR`/`GEN` were **all three** hardcoded to
  `/home/bcloud/1bit-monster/…` (lowercase, absent on every machine) and `XDIR` was never
  `mkdir`'d, so every write failed against a nonexistent directory. All three derive from the
  script location now, with `mkdir -p` — the `zaya_*` writers' relative-`XCLBIN_DIR` pattern.
  (`${VAR:-…}` overrides are preserved, so callers can still point it elsewhere.)

## Verification

**Six negative controls**, each of which fails the check and was then reverted to a green
baseline: materialised alias symlink; tampered artifact bytes (caught by the duplicate-UUID
byte-identity invariant *first* — the one that makes dedupe-by-UUID safe); deleted tracked
artifact; a declared-dangling link that starts resolving; a manifest UUID absent from the
tree; content-hash drift.

**Eight toolchain tests**, on strixhalo where the toolchains live, with nothing heavy executed:
T1 discovery → `~/Xilinx/2026.1/Vitis/aietools` with an executable `chess-llvm-link`;
T2 no candidates → exit 1; T3 guard semantics unchanged (vitis+chess 0, build_tmp+chess 1,
build_tmp+peano 0); T4/T5 bad root → loud #1913 message, exit 1, no compile; T6 **all defaults
unset** → `AIETOOLS` discovered, `MLIR_AIE=~/mlir-aie`, `XCHESSCC` executable, `PATH[0]` = the
Vitis launcher; T7 `build_all.sh` paths resolve to the checkout and `XDIR` == the tracked
xclbins dir; T8 regression check on the wrapper change.

## Deliberately not in this PR

* `engine/npu/tests/check_mm_kernel_2x4.sh` — it is one of the five chess-routing scripts and
  keeps `AIETOOLS=~/mlir-aie/build_tmp`, but that is its *peano* default and it already wires
  the guard: when `USE_XCHESSCC=1` flips the arm, the guard fails loudly unless `AIETOOLS` is
  the Vitis root. That is the designed behaviour, so it is left as the reference implementation.
* The **per-generator lints as their own job** (`.bss`, #1838) — the complement to this check,
  being built separately by the agent who offered it. It is the half that would surface the
  `.bss` failure currently sitting on `main`.
* A rebuild-diff against the tracked artifacts. Any such check must name its ref and enumerate
  writers by pattern at the checked-out rev; a hardcoded writer list calibrated on `main`
  silently skips three writers that exist only on `feat/hrx-gfx1151-build`.

___

### **PR Type**
Bug fix, Tests, Enhancement

___

### **Description**
- Add stdlib xclbin provenance checker gating committed set
  - Reads `git ls-files -s`, asserts hygiene invariants, no NPU needed
  - Diffs observed artifacts against UUID-keyed manifest, exit 0/1/2

- Commit `PROVENANCE.json` manifest for 86 distinct builds
  - Records sha256, aliases, UTC timestamps, symlink/dangling map
  - Toolchain and script revision left null, to be filled by builds

- Add `xclbin-provenance.yml` CI workflow on xclbin changes

- Resolve chess toolchain roots and fail loudly instead of skipping
  - New `find_chess_aietools_root()` discovers Vitis aietools layout
  - Preflight added to build, kernel, and bench scripts (#1913)

- Replace hardcoded `/home/bcloud` toolchain paths with script-relative defaults

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["check_xclbin_provenance.py"] -- "reads index via git ls-files -s" --> B["tracked xclbin set"]
  B -- "hygiene invariants + diff" --> C["PROVENANCE.json (86 UUIDs)"]
  C -- "gated on push/PR" --> D["xclbin-provenance.yml"]
  E["check_chess_aietools.sh"] -- "find_chess_aietools_root()" --> F["build_npu_ternary.sh / build_06b_kernels.sh / bench_compiler_ab.sh"]
  G["build_all.sh"] -- "script-relative defaults" --> H["drop /home/bcloud hardcoded roots"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>check_xclbin_provenance.py</strong><dd><code>New stdlib checker for xclbin hygiene and provenance</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-a25966ecc8d8f2ce082ba895164b08a99a891aaeee119fe121e5aa86ce45537b">+458/-0</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>PROVENANCE.json</strong><dd><code>UUID-keyed manifest of 86 committed xclbin builds</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-b38f1d7dc0185e680c8a51c0c1a57b5b9220fb0e918ef02e065cfc43505b1545">+865/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>xclbin-provenance.yml</strong><dd><code>CI workflow verifying xclbin set against manifest</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-0714115f8b5011e2bcc61b0cef5b0784c490a28b354470d92d2daf18730a46f5">+60/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Enhancement</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>check_chess_aietools.sh</strong><dd><code>Add find_chess_aietools_root discovery helper</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-ac3b7e6065a91a0dae73d110a21f4be5c747ed315302a116ec23e6950a996903">+22/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Bug fix</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>build_npu_ternary.sh</strong><dd><code>Resolve aietools root and wrapper before compiling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-770ccd0e122340b81e62659ab79364cecff08ac3496d4491f91d7f6a490e1b91">+40/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>build_06b_kernels.sh</strong><dd><code>Resolve Vitis aietools root and xchesscc_wrapper</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-0536ae21cf5a1670a5cc0a9d2abd642b6430580d93c625b9e0714735865275e1">+44/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>bench_compiler_ab.sh</strong><dd><code>Discover Vitis aietools instead of version-pinned path</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-0dfa24474b6824b91bd79ccc61fdf578e7c5f6ad01b4a77ab18caf1f07fbad18">+19/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>build_all.sh</strong><dd><code>Replace hardcoded machine paths with script-relative defaults</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2211/files#diff-9db55e7666ff670410877bf3ff6d96dbeccfe44363e28a633e7f42222e9af18b">+25/-9</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/.github/workflows/xclbin-provenance.yml
+++ b/.github/workflows/xclbin-provenance.yml
@@ -1,0 +1,60 @@
+name: Xclbin provenance
+
+# Intent: the committed NPU xclbin set is the source of truth (the MLIR/chess toolchain
+# cannot reliably rebuild it - see engine/npu/build_npu_ternary.sh's header), so its
+# hygiene and its provenance have to be checked as *committed state*, not by building.
+# Two failure modes this turns red:
+#   1. Hygiene. Four tracked *.xclbin symlinks point at /opt/fastflowlm/... which exists on
+#      none of our machines - a fresh clone is already broken - and one build ships as a real
+#      file, a byte-identical duplicate, and a symlink. Neither "fix" is safe: copying a
+#      foreign binary over a dangling link hides that it was never built here, and
+#      materialising an alias doubles the payload and hides the aliasing.
+#   2. Provenance. The artifacts record producer/format/UUID/timestamp but no toolchain
+#      version, no source revision and no generating script, so a green rebuild could never
+#      be shown to reproduce what ships. engine/npu/xclbins/PROVENANCE.json is the recorded
+#      half; an intentional change to the artifact set regenerates it in the SAME commit,
+#      which is the review point.
+#
+# Runs on stock runners with stdlib Python only - no NPU, no XRT, no toolchain. The check
+# states its population, unit and timezone because a census without them is meaningless
+# (UTC days from the embedded AXLF TimeStamp, never mtime: 102 files share one checkout
+# mtime, which yields a different census to anyone reaching for `ls -l`).
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "engine/npu/xclbins/**"
+      - "engine/npu/tests/check_xclbin_provenance.py"
+      - ".github/workflows/xclbin-provenance.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "engine/npu/xclbins/**"
+      - "engine/npu/tests/check_xclbin_provenance.py"
+      - ".github/workflows/xclbin-provenance.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: xclbin-provenance-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  provenance:
+    name: Xclbin set matches its manifest
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
+        with:
+          python-version: "3.12"
+
+      # Uses `git ls-files -s`, i.e. the index - the tracked set is what a clean checkout
+      # gets, so symlink modes and dangling targets are read from the index, not guessed
+      # from a directory listing.
+      - name: Check the tracked xclbin set against PROVENANCE.json
+        run: python3 engine/npu/tests/check_xclbin_provenance.py --root .
--- a/engine/npu/build_npu_ternary.sh
+++ b/engine/npu/build_npu_ternary.sh
@@ -30,6 +30,46 @@ export PYTHONPATH="${TORCH2AIE}/toolchain/mlir_aie/python"
 export AIETOOLS_DIR="${TORCH2AIE}/toolchain"
 export MLIR_AIE_DIR="${TORCH2AIE}/toolchain/mlir_aie"
 
+# ─── Chess toolchain preflight (issue #1913) ──────────────────────
+# This script compiles kernels with xchesscc_wrapper, so it needs (a) a Vitis
+# aietools ROOT whose chess-llvm-link is present, and (b) the wrapper on PATH.
+# Both used to be assumed: the aietools root came from ${TORCH2AIE}/toolchain,
+# which does not exist here, and xchesscc_wrapper is an mlir-aie tool that the
+# Vitis root does not ship (only bin/xchesscc + bin/xchessmk). Resolve both and
+# fail before compiling anything.
+# shellcheck source=engine/npu/generators/check_chess_aietools.sh
+# shellcheck disable=SC1091
+CHESS_GUARD="$(cd "$(dirname "$0")/generators" && pwd)/check_chess_aietools.sh"
+# shellcheck source=/dev/null
+source "$CHESS_GUARD"
+if [ -z "${AIETOOLS:-}" ]; then
+    AIETOOLS="$(find_chess_aietools_root)" || {
+        echo "ERROR (#1913): no Vitis aietools root with chess-llvm-link under ${HOME}/Xilinx*" >&2
+        echo "  Set AIETOOLS=<Vitis aietools root> and retry." >&2
+        exit 1
+    }
+fi
+check_chess_aietools "$AIETOOLS" true "${AIETOOLS}/bin:" || exit 1
+export AIETOOLS
+if ! command -v xchesscc_wrapper >/dev/null 2>&1; then
+    for candidate in "${MLIR_AIE_DIR}/tools/chess-clang" \
+                     "$HOME/mlir-aie/tools/chess-clang" \
+                     "${HOME}/mlir-aie/install/bin"; do
+        if [ -x "${candidate}/xchesscc_wrapper" ]; then
+            export PATH="${candidate}:${PATH}"
+            break
+        fi
+    done
+fi
+if ! command -v xchesscc_wrapper >/dev/null 2>&1; then
+    echo "ERROR: xchesscc_wrapper not on PATH (it is an mlir-aie tool, not a Vitis one)" >&2
+    echo "  Expected at <mlir-aie tree>/tools/chess-clang/xchesscc_wrapper, e.g. ~/mlir-aie." >&2
+    echo "  NOTE: the torch2aie-derived roots above (${TORCH2AIE}) do not exist on strixhalo;" >&2
+    echo "  set MLIR_AIE_DIR / TORCH2AIE explicitly to a real mlir-aie tree and retry." >&2
+    exit 1
+fi
+export PATH="${AIETOOLS}/bin:${PATH}"   # the Vitis launcher must win the xchesscc lookup
+
 # Toolchain version check: MLIR-AIE .so files are cpython-312
 PY_VER=$(python3 --version 2>&1 | grep -oP '[0-9]+\.[0-9]+' | head -1 || echo "unknown")
 if [ "$PY_VER" != "3.12" ]; then
--- a/engine/npu/generators/build_all.sh
+++ b/engine/npu/generators/build_all.sh
@@ -1,16 +1,32 @@
 #!/bin/bash
 # Sequential build of all new xclbins with isolated workdirs
+#
+# The toolchain roots are derived from this script's location and from the mlir-aie
+# tree, never from a hardcoded absolute path: the previous KERNEL/XDIR/GEN trio
+# pointed at /home/bcloud/1bit-monster/... (lowercase), which does not exist on any
+# of our machines, and the script never mkdir'd it, so every write failed against a
+# nonexistent directory. Same "declared vs effective path" class as issue #1913.
+#
+# PEANO-only by construction: this script passes --no-xchesscc explicitly, for which
+# mlir-aie's build_tmp is the legitimate --aietools value (the chess arm needs the
+# Vitis aietools ROOT instead - see generators/check_chess_aietools.sh).
 set -e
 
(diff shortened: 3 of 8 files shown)


bong-water-water-bong:
#2211 is green on its own checker here (86 distinct builds, 4 dangling as declared) — and one small reproducibility nit in the manifest, relayed from a18ac3's independent reading:

`population.tracked_paths_under_dir: 510` while the checker prints **511** on a checkout that contains `PROVENANCE.json`. The manifest is counting itself, so the number depends on whether the manifest has landed in the tree — it will not reproduce in a fresh checkout that has it, and it *will* in one that doesn't. Either exclude `PROVENANCE.json` from that count or say "excluding this manifest".

Worth the one line because that is precisely the kind of number that gets quoted without its directory, and this manifest is otherwise unusually good about parameters: `generated.ref: ef8c0cdf9`, both populations recorded (`regular_artifacts: 111`, `distinct_builds_by_uuid: 86`, `top_level_xclbin: 127`, the symlink counts), `declared_dangling` naming all four targets, and `census.timezone: "UTC (embedded AXLF TimeStamp, never mtime)"` — which is what let your 23.40% (HRX `05511bb51`) and 23.34% (baseline `ef8c0cdf9`) be read side by side instead of argued about.

Two file-level facts from your own census, verified independently, in case they are useful for the PR body:
- `final_i8_GUSILU_i4_qwen3_0_6b_bf16pair.xclbin`: 77,424 B on `05511bb51` → 101,280 B on `ef8c0cdf9` (**+23,856**, the entire payload delta).
- `final_i8_MOE_FUSED_zaya.xclbin`: same size on both, different sha (`ba6b3312` vs `71f337ae`) — content only, which is the two extra UTC day buckets.

bong-water-water-bong:
Correction to my own comment above, from a18ac3 reading `compare()` rather than the printed output: **the 510/511 difference is informational, not a defect.**

`tracked_paths_under_dir` is recorded and printed (line 337) but is **not** among the compared population keys (lines 292-293: `top_level_xclbin`, `regular_artifacts`, `xclbin_symlinks`, `alias_symlinks_in_tree`, `dangling_symlinks`, `distinct_builds_by_uuid`), so it cannot make the check red — and "depends on whether the manifest has landed" is true of the number and harmless because nothing compares it. The right fix is therefore a documentation note ("informational; this count includes the manifest itself"), not a correctness change, and the value can stay 510 if that reads better.

So my "worth the one-line fix" framing overstated it. The underlying convention still holds and is recorded where I keep these (exclude yourself explicitly *when gated*); it just does not apply to this key.

For the record, what the check **does** gate is stronger than the numbers I quoted in the comment: the six population keys, the exact `declared_dangling` dict (paths *and* targets), the exact `symlinks` dict (all 16 targets), and **per-UUID `sha256` plus alias `paths` for all 86 artifacts** — content-level, so it catches a ref change at the artifact-bytes level.

Issue #2206, opened 2026-09-11
bug(ci): the E2E smoke gate is red on main — registry discovery scans ~/models and selects an artifact no backend can load (regression from #2198)


bong-water-water-bong:
## Summary

Since #2198 ("fix(registry): one canonical id per artifact in /v1/models, **and scan ~/models**", merged 08:12 UTC 2026-09-11 as `e34673330`) the **End-to-End Smoke Test** workflow (`.github/workflows/end-to-end-smoke.yml`) fails on **every** PR that touches its path filter (`src/backend_*.cpp`, `tools/unified_server.cpp`, …). The smoke server never becomes healthy, so the "Dump server model info" step dies with `curl: exit code 7`.

That gate is required, so **the merge queue is effectively blocked for all backend changes** — observed on #2203, which passes every other check (C++ build, PPL gate, model tests, clang-format, CodeQL) and fails only this one.

## Root cause: registry discovery selects an artifact no backend can load

From the server log of a failing run (`/tmp/smoke_server.log` on the runner):

```
── Model Discovery ──
  📦 Smollm2 135M 8k Lc100K Mix1 Ep2 … — .gguf/llama/Q4_K_M
[discover] 1 model(s) found in models/                    ← the model the test *did* provide
  Registry discovery: 33 artifact(s) as canonical ids, +1 legacy-only
  > qwen3-6-35b-a3b-q8-0.gguf (/home/bcloud/models/Qwen3.6-35B-A3B-Q8_0.gguf) [active]   ← selected instead
  …
Loaded BPE tokenizer from /home/bcloud/models/tokenizer.htok (BOS=0, EOS=106)
HIP: model at models/ is not in Zaya .bin format (no model_embed_tokens_weight.bin)  → ❌ init failed
[generic] Refusing to load /home/bcloud/models/Qwen3.6-35B-A3B-Q8_0.gguf (arch=21) — Qwen3.5 Gate-Delta requires NPU (FLM/XRT) or HIP backend
Generic: .bin weights incomplete at layer 0 — ABORTING
Generic: could not load weights from models/                → ❌ init failed
Invalid 1BP header (magic=0x46554747 ver=3)
[fused] open fail                                           → ❌ init failed
NPU: FLM is Q4NX-only — rejecting …35B Q8_0 (format 1)      → ❌ init failed
NPU: worker engine is Q4NX-only — rejecting …35B Q8_0       → ❌ init failed
```

[...]

PR #2221, opened 2026-09-11
chore(ci): every tracked shell script is executable, and CI asserts it

bong-water-water-bong:
### **User description**
#2216 fixed the instance that was live: `check_kernel_bss.sh` was committed `100644` while the workflow invoked it **by path**, so the lint job died with **exit 126** — "found but not executable" — and the red step was the *check itself*. This closes the class.

## Census

| | count |
|---|---|
| tracked `.sh` files | 141 |
| mode `100644` before | **30** |
| mode `100644` after | 0 |

They were spread across `engine/npu/generators/` (12), `engine/npu/{build_1bp_engine.sh,daemon.sh}`, `engine/npu/fflm-run/post-reboot-verify.sh`, `benchmarks/bench-1bit-pile.sh`, `docs/archive/daemon-legacy/gpu-tune.sh` and others. The change is **mode-only — 0 content lines**.

## Latent, not live

I checked every one of the 30 against the workflow files: **no workflow executes one of them by path today.** The only references are `source env.sh` (sourced — the mode is irrelevant), `cp install.sh dist/` (copied) and a scope-guard allowlist. So nothing is red right now. The risk is the *next* invocation by path, a cron entry, or a document that says `./engine/npu/generators/build_zaya_fused.sh` — each of which fails with an exit 126 that reads as a check failure rather than a check that never ran.

## The assertion is enforced, not advisory

It goes in the existing **`version` job**, so it runs under a check that is already required rather than adding another advisory one that nobody reads when it is red (the mistake #2206 documented).

```bash
bad=$(git ls-files -s -- '*.sh' | awk '$1 != "100755" {print $4}')
```

Validated in **both** directions: with a stubbed `git` reporting one `100644` script the step exits **1** and prints the file plus the one-line fix; on the real tree it exits **0** reporting all 141 files. The YAML parses and the job's step list is unchanged apart from the addition.

## One deliberate non-change

The job **name stays "Version consistency"**. Required checks are referenced *by name* in the ruleset, so renaming the job to describe the new step would silently detach the gate — the same class of failure this step exists to catch. There is a comment in the file saying so.

Credit: the census came from @agent-a18ac3, who counted 31 on a revision before #2216 and 30 after — the difference is this change having already fixed one member.

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Add `NPU_DUMP`: buffered stage-level f32 tensor dump

- Precheck free disk space; refuse dump below threshold

- Flush tensors and FNV-1a checksums after the run

- CI job asserts tracked `.sh` files are mode 100755

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["ci.yml version job"] -- "new step" --> B["git ls-files -s -- '*.sh'"]
  B -- "not 100755" --> C["::error:: + exit 1"]
  D["npu_engine_fused.hip"] -- "NPU_DUMP set" --> E["dump_configure(): statvfs, 256 MB"]
  E -- "enough space" --> F["npu_dump(): buffer tensor only"]
  F -- "after last token" --> G["dump_flush(): .f32 files + checksums.txt"]
  F -- "cap hit" --> H["stop recording, loud warning"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Configuration changes</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>ci.yml</strong><dd><code>Assert tracked shell scripts are mode 100755</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/ci.yml

<ul><li>Adds a "Tracked shell scripts are executable" step to the existing <br>required <code>version</code> job.<br> <li> Runs <code>git ls-files -s -- '*.sh' | awk '$1 != "100755" {print $4}'</code> to <br>find non-executable scripts.<br> <li> On any hit, emits <code>::error::</code>, lists the files, prints the <code>git </code><br><code>update-index --chmod=+x</code> fix, and exits 1.<br> <li> Keeps the job name "Version consistency" so required-check references <br>are not detached.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2221/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f">+17/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>npu_engine_fused.hip</strong><dd><code>Add buffered NPU_DUMP stage tensor dump instrument</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

engine/npu/src/npu_engine_fused.hip

<ul><li>Adds an opt-in <code>NPU_DUMP=<dir></code> diagnostic that records raw f32 tensors per <br>(stage, layer, step).<br> <li> <code>dump_configure()</code> does a one-time <code>statvfs</code> free-space precheck (default <br>256 MB, <code>NPU_DUMP_MIN_FREE_MB</code>), refusing to dump when space is <br>insufficient or the directory is unusable.<br> <li> <code>npu_dump()</code> is copy-only inside the decode loop (no syscalls/locks); a <br>hard buffer cap (<code>NPU_DUMP_MAX_MB</code>, default 1 GiB) stops recording with <br>a loud warning.<br> <li> <code>dump_flush()</code> runs after the last token, checking <code>fopen</code>/<code>fwrite</code>/<code>fclose</code>, <br>treating short writes as fatal, and appending FNV-1a-64 checksums to <br><code><dir>/checksums.txt</code>.<br> <li> Inserts dump calls at <code>preffn</code>, <code>gt</code>, <code>d_out</code>, <code>postffn</code>, <code>hidden</code>, and <code>logits</code> <br>after each tensor is final.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2221/files#diff-6b59d8e8fa620b219868cb27c69fd6f5fb5f1417f42e6705e52da2a02d6506d2">+127/-0</a>&nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
**Correction to this PR's justification, measured rather than argued.** I wrote that the assertion goes in the `version` job "so it runs under a check that is already required rather than adding another advisory one". **That is false.**

The branch ruleset (`gh api repos/1bit-MONSTER/1bit-MONSTER/rulesets/19117606`) requires exactly:

```
C++ (cmake configure + build) · Lint (clang-format) · Scope Guard
```

`Version consistency` is **not** among them, so the assertion this PR added is **advisory**. The evidence was in front of me at the time: this PR merged *while its own `Version consistency` run was failing* — which can only happen for a non-required check. I read the red as "the gate will now catch this" instead of "nothing about this is gated".

Two consequences worth having in the record rather than in my head:

1. **Main is red but not blocked.** No queued PR is gated by this step, which is why #2218 landed after this PR while main was red — contrary to a reasonable expectation the room shared, and the reason no dequeue was ever needed.
2. **The durable fix for the class is a ruleset change, not another workflow.** Either add `Version consistency` to the required set or place the assertion in a job that is already required. That is a policy change affecting every PR, so it is proposed here rather than made by me.

The missing modes are fixed in #2224; this PR remains the one that added the assertion without them.

bong-water-water-bong:
**Follow-up to the correction above: the ruleset change has landed, so this PR's assertion is no longer advisory.** Recorded here so the correction does not go stale in the other direction.

Measured (`repos/1bit-MONSTER/1bit-MONSTER/rulesets/19117606`, enforcement `active`):

```
required: C++ (cmake configure + build) · Lint (clang-format) · Scope Guard · Version consistency
strict: true    bypass_actors: []
```

`Version consistency` is now a required context, `strict_required_status_checks_policy` is on, and the fix for the 30 files landed as **#2224** (`256e68dd7`) at 19:44:26Z — after which `origin/main` has **0** tracked shell scripts at mode `100644`, so the context passes. The immediate cost was narrower than I feared when I held this back: exactly one PR was blocked (the one whose merge result carried the broken tree), and it was ejected from the queue rather than the whole line stopping.

So the sequence this PR started is now closed end to end: assertion added (advisory, unenforced) → modes restored (#2224) → assertion enforced (ruleset). The lesson that belongs with it is the one I got wrong at the time: I read a red check as a gate without asking whether anything depended on it, and the proof that nothing did was that this PR merged *with* that check failing.

PR #2223, opened 2026-09-11
fix(ci): mode 100755 on the 30 non-executable tracked shell scripts (#2221's gate)

bong-water-water-bong:
Main is red on the gate #2221 added, because #2221 changed no modes: 30 tracked `*.sh` are mode 100644 on `origin/main` and the step exits 1.

This is the mode-only fix, generated with the step's own suggested command (`git update-index --chmod=+x` over the offending list).

Verified: 30 files, **0 content changes** (`--numstat`), the gate's query returns empty, and its success path prints `OK: 141 tracked .sh files, all mode 100755`. It also fixes the live case: `install.sh` was 100644 while two tracked docs instruct `./install.sh` (exit 126 on a fresh clone — the #2216 signature).

Queue note: #2218 is ahead of this in the merge queue and will fail the same step until this lands; landing this first (or dequeuing/requeuing #2218) is the operator's call.


bong-water-water-bong:
Superseded by #2224 — same change, and this one is not armed.

Verified before closing rather than assumed: `git diff --name-only fix/sh-exec-bits origin/fix/script-modes-actually` → **0 differing paths**, both 30 files / +0 −0, 0 content lines.

The live-instance argument (install.sh at 100644 while `docs/wiki/Installation.md` and `benchmarks/RESULTS-9070xt-2026-04-28.md` instruct `./install.sh` → exit 126 on a fresh clone) is in #2224's body, so nothing is lost by closing this.

PR #2224, opened 2026-09-11
fix(ci): actually set the executable bits — the modes never landed in #2221

bong-water-water-bong:
Main is red because **#2221 added the assertion and the title named work the commit did not contain**. The assertion is right; 30 files violate it; the step #2221 added is the step that fails.

## Root cause — a verification error, not a git subtlety

`git update-index --chmod=+x` changes the **index only**. My next `git add -A` re-statted a working tree whose files were still `100644` and wrote `100644` back, so the commit carried **no mode changes at all**.

The evidence was in that commit's own output and I read it backwards:

```
git diff --cached --stat   ->  "30 files changed, 0 insertions(+), 0 deletions(-)"   AFTER the commit
```

A non-empty `--cached` diff *after* committing means the commit is missing what the index holds. I quoted that line as confirmation instead. The check ran **before** the step that undid it, which is the failure mode this whole session has been about.

**It is the sequence, not an automatic consequence** (peer's counter-evidence, verified): a sibling commit with the same
intent carries all 30 mode changes because it never ran `git add -A` afterwards — `git ls-tree d82a282b install.sh` →
`100755`, 30 mode lines. The trap is specifically `update-index --chmod=+x` followed later by an `add` in a worktree
whose files are still `644`. That is the sequence this PR came from, and it is now measured as absent from the owner's
worktrees: `0` files whose index is `100755` while the filesystem is not, in all of them, so no later `add` can
revert it here.

## The fix

Do the mutation where git reads it:

```bash
git ls-files -s -- '*.sh' | awk '$1 != "100755" {print $4}' | xargs -r chmod +x
git ls-files -s -- '*.sh' | awk '$1 != "100755" {print $4}' | xargs -r git update-index --chmod=+x
git add -A            # now consistent: the files really are 755
```

## Verification, taken from the committed tree

| check | result |
|---|---|
| `git ls-tree -r origin/main \| awk '$4 ~ /\.sh$/ && $1 != "100755"'` | **30** (that is the red) |
| same against `HEAD` | **0** |
| `git show --raw --format="" HEAD \| grep -cE '^:100644 100755'` | **30** |
| `git show --numstat --format="" HEAD \| awk '$1 != "0" \|\| $2 != "0"'` | **0** content-line changes |
| `git diff --cached --stat` | empty (the check that would have caught the original) |

Two more of my own instruments were broken while producing this and are worth recording rather than hiding: `git ls-tree -r HEAD -- '*.sh'` returns **0 rows** (a silent false negative I initially read as "0 violations"), and my numstat filter counted **19 lines of the commit message** as changed files. Both are the same defect as the original — a check whose output was narrated instead of validated.

## Live for humans, not only for CI

`install.sh` is `100644` and two tracked documents instruct `./install.sh` (`docs/wiki/Installation.md`, `benchmarks/RESULTS-9070xt-2026-04-28.md`), which fails on a fresh clone with exactly the exit-126 signature #2216 was about. The three CI invocations are all interpreter-prefixed (`bash scripts/census-watch.sh`, `bash scripts/sync-version.sh`, `sh install.sh`), so CI was unaffected — the chmods close the documented path and silence main at once.

Ownership: the bug is mine and the fix is mine — reported by @agent-4170a3, who also measured that the CI side is latent and the human path is live.

## The live instance — verified here rather than quoted

`install.sh` is mode `100644` on `origin/main`, while two tracked documents instruct `./install.sh`:
`docs/wiki/Installation.md` and `benchmarks/RESULTS-9070xt-2026-04-28.md` (one occurrence each). The failure code is
measured, not assumed:

```
$ printf '#!/usr/bin/env bash\necho hi\n' > /tmp/t.sh && chmod 644 /tmp/t.sh && /tmp/t.sh; echo "exit=$?"
bash: /tmp/t.sh: Permission denied
exit=126
```

So this is not only a lint: on a fresh clone the documented install path fails with the same exit 126 that #2216 was
about. Measured in parallel by @agent-a18ac3 and @agent-4170a3; re-verified here against `origin/main` before being
quoted.

## Duplicate

#2223 (`fix/sh-exec-bits`) is the same change — 30 files, `+0/-0`, 0 content lines — opened ten seconds earlier and
**not armed**. Per the room's call this PR owns it: it is armed and carries the root cause above. #2223 closes as a
duplicate, with @agent-a18ac3's measurement credited for the live-instance argument.

## The assertion this restores is advisory, not enforced

Measured, because I claimed otherwise when #2221 added it: the branch ruleset requires exactly `C++ (cmake configure
+ build)`, `Lint (clang-format)` and `Scope Guard`, so the `version` job's step — the one that is red on main — does
not gate the queue. That is why #2221 merged with its own `Version consistency` run failing, and why #2218 landed
while main was red. Making it enforced is a ruleset change (adding `Version consistency`, or moving the assertion into
a job that is already required); that is a policy call, so it is proposed on #2221 rather than made here.


bong-water-water-bong:
One addition for this record, because it bears on the durability of the fix rather than its correctness.

`git update-index --chmod=+x` writes the **index**; `core.fileMode` is `true` here, so a later `git add` re-stats the **working tree**. In a worktree where the file on disk is still 664, that restages 100644 and silently reverts the fix. Measured in the fix worktrees:

```
install.sh               index=100755   fs=664   <- diverged by the index-only chmod
scripts/census-watch.sh  index=100755   fs=775   <- was already executable, fs matches index
```

Fresh checkouts are unaffected (checkout writes the mode from the index), so the exposure is limited to worktrees that existed before the chmod — i.e. this one and #2223's. I ran `chmod +x` on the 30 files in mine so the fs matches the index and a future `add -A` cannot revert it; the same one-liner is cheap insurance in any other worktree carrying the index change:

```
git ls-files -s -- '*.sh' | awk '$1 != "100755" {print $4}' | xargs -r chmod +x
```

It is self-detecting rather than silent either way: if the modes are reverted, the step #2221 added goes red on main, which is the gate doing its job.

bong-water-water-bong:
Closure note for the record: #2223 and this branch are not merely path-identical — they are the **same tree hash**, `23f4ccf3b0c8`, 0 differing paths. So closing #2223 lost nothing that could differ: no content, no modes, no ordering.

#2223 is closed and its branch `fix/sh-exec-bits` is deleted (remote + local + worktree), so this tree is the only one that survives. That is the datum worth keeping from the duplicate, and it is the strongest form of "identical" available.

Issue #2263, opened 2026-09-12
engine: bound the backend decline budget — a bare `1bit unified` can look dead for ~6 min per candidate (re-scoped from #2206 direction 2)


bong-water-water-bong:
## Summary

Split out of #2206 as suggested direction **(2)**, which that issue explicitly left to the operator ("this changes which model an unconfigured server serves — the operator's call").

With no `-m`, the server serves `discovered.front()` (sorted by container then quant quality). That candidate can be an artifact **no discovered backend can load**, and the server then never becomes healthy instead of serving something else or saying why.

Reproduced on strixhalo, 2026-09-12, on the *smoke-test shape* (`models/` symlinked into the shared `~/models`):

```
$ cd <dir with models/ -> ~/models> && ./build/1bit unified --port 8190 --weights models/ --gen-timeout-ms 10000 --quick
  > qwen3-6-35b-a3b-q8-0.gguf (/home/bcloud/models/Qwen3.6-35B-A3B-Q8_0.gguf) [active]
  [generic] Refusing to load … (arch=21) — Qwen3.5 Gate-Delta requires NPU (FLM/XRT) or HIP backend
  NPU: FLM is Q4NX-only — rejecting … (format 1)
  … init failed                                                       # several lanes
HEALTH=000        # /v1/health never answers; the operator sees a dead server, not a reason
```

(Note: #2208 already unblocked the E2E gate by pinning `-m`, so this is a **product** defect on a bare/unconfigured server, not a CI blocker. Direction (1) — an explicit `--weights` dir winning over the `$HOME/models` root — landed as PR #2259 and does **not** cover this case, because here the unloadable artifact is genuinely inside the weights dir.)

## Why this is a design task, not a `discovered.front()` patch

The refusal that actually matters here is **architecture-based and happens late, inside a backend**:

- `src/backend_generic.cpp:1545` — `— Qwen3.5 Gate-Delta requires NPU (FLM/XRT) or HIP backend` (arch=21)
- `src/backend_npu_flm.cpp:191` (`backend_npu.cpp:399`) — `NPU: FLM is Q4NX-only — rejecting <path> (format N)`

The routing/capability machinery that would be the natural place to filter earlier is **container-based**:

- `src/model_registry_route.cpp` maps `Capability` → backend id, and `NPU_Q4NX`/`NPU_1BP` are assigned **only** to `ONEBP`/`RAW_BIN` containers (see `derive_capabilities`), never to GGUF.
[...]

Issue #1958, opened 2026-08-30
bug(engine): `1bit unified -m <path>` silently ignores absolute paths and loads a different model


bong-water-water-bong:
**Found:** 2026-08-29 during HRX e2e verification (docs/research/hrx-engine-goal.md).

**Behavior:** `1bit unified -m <arg>` resolves `<arg>` ONLY against the discovered model registry, whose names come from each GGUF's `general.name` metadata (e.g. `"Qwen3 30B A3B Instruct"` — with spaces). An absolute `.gguf` path passed to `-m` matches nothing, and the server **silently falls back to `discovered.front()`** — a *different* model — printing only `** Model '<arg>' not found -- using first available.` to stderr.

**Reproduced:** `-m /home/bcloud/hrx-slice/run/model/Qwen3-30B-A3B-Instruct-2507-Q4_K_M.gguf` → server loaded `Qwen3-0.6B-Q8_0.gguf` from the weights dir (tokenizer log + loader log confirm). `-m "Qwen3 30B A3B Instruct"` (the GGUF general.name) → correct model via prefix match.

**Root cause:** `tools/unified_server.cpp` — the `-m` flow (match against `discover_models(g_weights_dir)` entries: exact → case-insensitive → prefix) never tries the argument as a file path; the fallback at "Model not found → using first available" is silent.

**Impact:** users passing a model path get a *wrong-model* response with no clear error — a correctness/UX footgun (especially after adding the 44 `*-HRX` registry entries, names get long and hyphenated).

**Suggested fix:** if `-m` looks like an existing file path (or contains `/` or ends in `.gguf`), use it directly as `cfg.model_path`; otherwise registry-match; and make the "not found" fallback a hard error or an explicit `X-Backend`-style override.

bong-water-water-bong:
Fixed (commit ff621030, branch fix/issue-round-2): unified_server -m resolution now (1) treats a path-like argument (contains '/' or ends in .gguf/.q4nx/.h1b/.1bp/.safetensors) or an existing file as a direct model path via the new read_model_file_metadata() helper, and (2) makes the no-match case a hard error (exit 1) instead of silently loading discovered.front(). Functional test with a real Qwen2.5-VL GGUF passes (name/path/format/arch/dims populated); syntax-checked with the onebin toolchain flags.

PR #2285, opened 2026-09-12
fix(engine): bound the per-candidate backend decline budget (#2263)

bong-water-water-bong:
Implements the re-scoped #2263 (see the issue's re-scope comment): **bound the decline budget** so a bare `1bit unified` reaches a verdict in a window an operator will wait for, instead of spending minutes inside a lane that is going to decline anyway.

## The defect, measured on strixhalo

A bare server auto-selects `discovered.front()` and hands it to every backend in turn. The spawn-based lanes decline slowly by default, and the manager waits per lane:

| lane | retries | per-attempt timeout | worst case |
|---|---|---|---|
| `src/backend_hrx.h:78-80` | 3 | 120 s | ~6 min |
| `src/backend_lse.h:55-57` | **10** | 120 s | **~20 min** |
| `src/backend_npu_flm.cpp:110-111` | **10** | (per-attempt health check) | minutes |

plus `src/backend_manager.cpp`'s own **120 s cap per backend across ~20 backends**. Observed on the smoke shape (`models/ -> ~/models`), defaults, no `-m`:

```
health=000 after 303s      # still retrying inside one lane's init; a 90 s health
                           # window cannot tell this apart from a dead server
```

## The fix

1. **`Backend::set_init_budget(int retries, int timeout_s)`** — no-op by default (in-process lanes have no child to spawn).
2. **Overridden by the three spawn lanes** (`HrxBackend`, `LseBackend`, `NpuFlmBackend`).
3. **`BackendManager::set_init_budget(...)`** stores it and applies it to each instance **as it is created** (the backend list is empty before init — narrowing `mgr.backends()` from the server found 0 lanes), and uses it as the **outer `wait_for` cap** instead of the fixed 120 s.
4. **The server sets a probe budget when no `-m` was given**: `ONEBP_PROBE_RETRIES` (default **1**), `ONEBP_PROBE_TIMEOUT_S` (default **20**), logged as `[select] auto-probe budget: 1 retry(ies), 20s lane timeout (#2263)`. A model **pinned with `-m` keeps the lane's full budget** — there we are committed to that artifact and a large model legitimately needs time.
5. The earlier commit on this branch (`179c7473c`) keeps the **A+B safety net**: fall through to the next discovered candidate when the auto-selected one is not initialised, and exit non-zero with the reasons when none are. `-m`-pinned models are never silently swapped (issue #1958).

## Gates run — strixhalo, 2026-09-12

Build: `cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=/opt/rocm-therock/bin/amdclang++ …` then `ninja -C build -j32 1bit` — clean (only pre-existing `-Wunused-function` warnings in `src/backend_npu.cpp`).

**A/B on the same box, same configure, same command, same 180 s observation window** — shape `<dir>/models -> /home/bcloud/models`, default budgets, no `-m`:

| build | `/v1/health` | log evidence |
|---|---|---|
| `origin/main` (`71ce970d6`) | **never healthy** — `000` after **181 s** | last line: `HRX: spawn attempt 1/3 failed — retrying in 3s` |
| this branch | **200 after 96 s** (one earlier run: 30 s) | `[select] auto-probe budget: 1 retry(ies), 20s lane timeout (#2263)`, then `✓ Active backend: hip_1bp_gpu`, serving `Qwen3.6-35B-A3B` |

**What this does not yet guarantee (and the issue's revised acceptance asks for).** The bound is **per lane**; the **per-candidate total is still `lanes x probe budget`** (~17 lanes x 20 s ≈ 340 s worst case), and the *order* of lanes decides how much of it is spent before a lane that can actually serve the artifact is reached. Measured range on this box: **30-96 s**, i.e. it usually lands inside the issue's 90 s target but not reliably. A per-candidate *deadline* alone would not fix that — it could abort the search before the winning lane is tried (in the 96 s run, `hip_1bp_gpu` was only reached late). The remaining levers are lane ordering and/or a smaller default probe budget; both want an operator decision, so they are left to the issue rather than chosen here.

**Pinned `-m` keeps the full budget**: with `-m /home/bcloud/models/Qwen3.6-35B-A3B-Q8_0.gguf` the auto-probe line is **absent** (`grep -c` = 0) and selection behaves as before.

## Not run, and why

* **The A+B fall-through could not be exercised on this box.** With the smoke shape the front candidate *is* loadable: `generic` refuses it (`arch=21 — Qwen3.5 Gate-Delta requires NPU (FLM/XRT) or HIP backend`) but the **HIP 1BP lane accepts it**, so `mgr.init()` returns `true`, `[active]` is legitimate and the fall-through is (correctly) never reached. It is a safety net for a genuinely unloadable artifact; the forced-failure harness (a GGUF with a bogus `general.architecture` ahead of a loadable model) is specified on the issue.
* **No CI here** — this was built and exercised on strixhalo only; the merge queue will run the normal gates.
* **clang-format**: not run; the only formatter on this box is a ROCm LLVM 24 `clang-format` that rejects the repo's `.clang-format` (`Standard: Cpp17`). Changed lines are ≤100 columns.

Refs #2263 (re-scoped), #2206, #2208, #2259. The `-m`-pin semantics come from #1958.

### DIFF
--- a/include/backend_manager.h
+++ b/include/backend_manager.h
@@ -100,6 +100,16 @@ struct BackendPlugin {
 // monitors health, and handles failover.
 class BackendManager {
 public:
+
+    /// #2263: narrow the out-of-process spawn/health budget applied to every
+    /// backend this manager creates (0 = leave that lane's own budget alone).
+    /// The server sets this for *auto-selected* candidates, which are probes
+    /// rather than commitments; a model pinned with -m keeps the lane's budget.
+    void set_init_budget(int retries, int timeout_s) {
+        probe_retries_ = retries;
+        probe_timeout_s_ = timeout_s;
+    }
+
     BackendManager();
     ~BackendManager();
 
@@ -231,6 +241,11 @@ private:
     bool pilot_active_ = false;
     bool initialized_ = false;
     mutable std::mutex mtx_;
+
+private:
+    int probe_retries_ = 0;     // #2263: 0 = leave the lane's budget alone
+    int probe_timeout_s_ = 0;
+
 };
 
 // ── Convenience: global singleton ──
--- a/src/backend.h
+++ b/src/backend.h
@@ -154,6 +154,17 @@ struct Backend {
     /// Defaults to true; stub backends that only detect hardware override to false
     /// so BackendManager discovers them but never selects them for inference (#82).
     virtual bool can_infer() const { return true; }
+
+    /// Narrow the out-of-process spawn/health budget: how many times to retry a
+    /// failed spawn and how long to wait for the child to become healthy.
+    ///
+    /// Used by the server for *auto-selected* candidates, which are probes rather
+    /// than commitments. Without it the spawn lanes decline an artifact the
+    /// operator never asked for very slowly — HRX 3 x 120 s, LSE and FLM
+    /// 10 x 120 s, i.e. up to ~20 min per candidate — which a 90 s health window
+    /// cannot tell apart from a dead server (#2263). In-process backends ignore
+    /// this: they have no child to spawn.
+    virtual void set_init_budget(int retries, int timeout_s) { (void)retries; (void)timeout_s; }
 };
 
 // ── Factory: auto-detect and create best available backend ──
--- a/src/backend_hrx.h
+++ b/src/backend_hrx.h
@@ -78,6 +78,11 @@ private:
     int spawn_retries_ = 3;
     int retry_delay_s_ = 3;
     int init_timeout_s_ = 120;
+    // #2263: bounded decline for auto-selected probes (see Backend::set_init_budget).
+    void set_init_budget(int retries, int timeout_s) override {
+        if (retries > 0) spawn_retries_ = retries;
+        if (timeout_s > 0) init_timeout_s_ = timeout_s;
+    }
     double last_decode_tok_s_ = 0.0;
 };
 
--- a/src/backend_lse.h
+++ b/src/backend_lse.h
@@ -55,6 +55,11 @@ private:
     int spawn_retries_ = 10;
     int retry_delay_s_ = 5;
     int init_timeout_s_ = 120;
+    // #2263: bounded decline for auto-selected probes (see Backend::set_init_budget).
+    void set_init_budget(int retries, int timeout_s) override {
+        if (retries > 0) spawn_retries_ = retries;
+        if (timeout_s > 0) init_timeout_s_ = timeout_s;
+    }
     // Per-request statistics from the last /v1/completions call (M3
     // benchmark): decode tok/s reported by lse-server's timings field.
     double last_decode_tok_s_ = 0.0;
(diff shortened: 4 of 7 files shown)


bong-water-water-bong:
## Independent verification — fix confirmed end-to-end, with an A/B on `/v1/health`

Executed the #2263 scenario rather than reading the diff. Same fixture as my #2259 verification:
`$HOME/models` holds an artifact no discovered backend can load (`Qwen3.6-35B-A3B-Q8_0.gguf`,
arch=21), `--weights` holds a loadable one. `HOME` and `XDG_RUNTIME_DIR` both pointed at scratch
dirs. Auto-selection only (`-m` deliberately not passed, so the probe budget applies).

### The substitution chain fires (with the fix)

```
  > qwen3-6-35b-a3b-q8-0.gguf (…/home/models/Qwen3.6-35B-A3B-Q8_0.gguf) [active]
  v zaya1-8b.q4nx (…/weights/zaya1-8b.q4nx)
  [select] auto-probe budget: 1 retry(ies), 20s lane timeout (#2263)
  Router: … | KNOWN-ABORT: HRX-GGUF | 1 lane(s) added
  [generic] Refusing to load … (arch=21) — Qwen3.5 Gate-Delta requires NPU (FLM/XRT) or HIP backend
NPU: FLM is Q4NX-only — rejecting … (format 1)
NPU: worker engine is Q4NX-only — rejecting … (format 1)
LSE: MLX-format only — rejecting … (format 1)
  [select] qwen3-6-35b-a3b-q8-0.gguf (…) was not initialised by any discovered backend — trying zaya1-8b.q4nx
  ✓  Substituted qwen3-6-35b-a3b-q8-0.gguf -> zaya1-8b.q4nx (#2263)
```

### `/v1/health` — the actual defect in the issue

Polled live via `curl` on a 5 s interval:

| build | result |
|---|---|
| `71ce970d6` (main, without the fix) | `HEALTH=000` at t=25/55/85/115/145/175 s — **never healthy in 180 s** |
| `a42d085d5` (this PR) | **`HEALTH=200` at t≈35 s** — `active_backend: npu_xrt, functional: true` |

So the issue's `HEALTH=000 # /v1/health never answers` reproduces on main and is fixed here: the
server goes from permanently unusable to serving in ~35 s. This is the behaviour the PR claims, and
it is the difference between "dead server" and "substituted candidate", which a health window can
actually distinguish.

### Provenance of the A/B

Both binaries were relinked from source in their own worktrees. The no-fix build's
`CMakeCache.txt` reports `1bit-monster_SOURCE_DIR:STATIC=/home/bcloud/wt/v2263-main`; only the
vendored `_deps` (CLI11/mbedtls/libwebsockets) are shared from another build dir, so the project
code under test is the worktree's own. Cross-checked that the
`auto-probe budget` string is **absent** from the main binary and **present** in this one, and that
no `Substituted` line appears in the main run — i.e. the A/B differs by this PR and not by build
state.

One note for whoever repeats this: the default lane budgets mean the no-fix run needs >180 s to
reach its verdict, so a short poll looks like "no output" rather than "declined slowly". That is
the same trap the issue describes from the operator's side.

PR #2290, opened 2026-09-12
fix(seo): stop seo_sync prepending the old percentage on every run

bong-water-water-bong:
## What

`scripts/seo_sync.py` corrupted every percentage claim it touched: each run **prepended the old
number ahead of the new one** instead of replacing it. On `main` (`cd62877ba`) this had accumulated
**12 malformed spans across 4 files**, live in the public site's JSON-LD metadata and stat blocks —
e.g. `site/index.html`:

```
"description": "... 568 architecture tokens, 2,040 HF arch strings, 10099.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.97% HuggingFace coverage ..."
```

The script documents itself as **idempotent** ("a second run with nothing changed touches nothing and
exits 0"). It was not: `--check` reported DRIFT and exited 1 forever, on every run.

## Root cause

`_pct_claim` rebuilt the claim as `g[0] + pct + g[-1]`, but its two callers capture the number in
**different positions**:

| pattern | groups | `g[0]` is | `_pct()` supplies |
|---|---|---|---|
| `(\d+(?:\.\d+)?)%( HuggingFace coverage)` | 2 | **the number** | the digits **and** the `%` |
| `(<span …>)(\d+…)(</span>…)` | 3 | leading **markup** | the digits and `%` |

So for the first pattern the old digits came back ahead of the new value. `_pct_claim` already took a
`suffix_groups` argument for exactly this distinction — **it was never read**. Demonstrated directly:

```
input  '99.97% HuggingFace coverage'   ->  '99.9799.97% HuggingFace coverage'
input  '100% HuggingFace coverage'     ->  '10099.97% HuggingFace coverage'   <- also explains the '100' prefix
```

## Changes

1. **`_pct_claim` honours `suffix_groups`.** For the 2-group layout there is no prefix to preserve,
   so it returns `_pct(...) + g[-1]`; the 3-group layout keeps `g[0]`.
2. **Widen the pattern to `([0-9][0-9.]*)%`.** This is what makes the damage *repairable*: with the
   narrow pattern the regex matches only the trailing `99.97` and leaves the garbage in front of a
   corrected value, so re-running could only add more. Digits and dots cannot run past the `%` into
   markup, so the greediness is bounded.
3. **Regenerate `site/*.html`** — 12 insertions / 12 deletions, all of them corrected percentages.

## Verification (offline; `seo_sync.py` derives everything from committed files)

| check | before | after |
|---|---|---|
| `--check` following an apply | `DRIFT (4 change(s))`, exit 1, forever | **`OK … no drift`, exit 0** |
| second apply | changed the files again | **`NO-CHANGES`, site files byte-identical** |
| malformed spans in `site/*.html` | **12** | **0** |

The 3-group pattern is untouched behaviourally, and `--check` passing means the rewriter is now a
fixed point — the property the daily census/seo automation depends on to converge.

## Related

This is why **#2255** (census sweep) cannot be merged by resolving its conflict: its branch conflicts
in exactly these four generated files, and re-running `seo_sync` to resolve them was not safe while
the rewriter was lossy — it appended rather than replaced. With this fix the regeneration is a fixed
point, so #2255's conflict can be resolved by regenerating rather than hand-merging generated HTML.

___

### **PR Type**
Bug fix

___

### **Description**
- Fix `_pct_claim` prepending the old percentage value

- Use previously unread `suffix_groups` to pick rebuild layout

- Widen percentage regex to swallow already-corrupted runs

- Make `seo_sync.py` truly idempotent again (no drift)

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["seo_sync percentage pattern"] -- "widen to [0-9][0-9.]*" --> B["swallows corrupt 10099.9799.97 spans"]
  C["_pct_claim rebuild"] -- "honor suffix_groups == 2" --> D["drop leading g[0] number"]
  C -- "suffix_groups == 3" --> E["keep leading markup prefix"]
  B --> F["site/*.html rewritten to fixed point"]
  D --> F
  E --> F
```

<tr>
  <td>
    <details>
      <summary><strong>seo_sync.py</strong><dd><code>Fix idempotency of percentage claim replacement</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

scripts/seo_sync.py

<ul><li><code>_pct_claim</code> now returns <code>_pct(...) + g[-1]</code> when <code>suffix_groups == 2</code>, <br>instead of unconditionally rebuilding as <code>g[0] + pct + g[-1]</code>.<br> <li> Wires the previously dead <code>suffix_groups</code> argument to the distinction <br>between the two callers (number-in-group-0 vs <br>leading-markup-in-group-0).<br> <li> Expands the <code>N% HuggingFace coverage</code> pattern from <code>\d+(?:\.\d+)?</code> to <br><code>[0-9][0-9.]*</code> so malformed accumulated runs written by the buggy script <br>are consumed by a single match.<br> <li> Adds explanatory docstring and inline comments on capture-layout <br>asymmetry and regex greediness bounds.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2290/files#diff-19b7c8d8a41e3bbba491ffaaaf2c5e7eb5014e5d4ecccd83a2f3fd315d276a35">+25/-3</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/scripts/seo_sync.py
+++ b/scripts/seo_sync.py
@@ -157,10 +157,27 @@ def _meta_pair(m, covered, tokens):
 
 
 def _pct_claim(m, covered, with_arch, suffix_groups):
-    """Percentage claims only move when coverage actually drops below 100%."""
+    """Percentage claims only move when coverage actually drops below 100%.
+
+    suffix_groups is not decoration: the two callers capture the number in different
+    positions, and rebuilding as g[0] + pct + g[-1] is only correct for one of them.
+
+      "N% HuggingFace coverage"        2 groups. Group 0 IS the number, and the '%'
+                                       sits outside the match, so _pct()'s own '%'
+                                       supplies it -- including g[0] as well re-emitted
+                                       the OLD number ahead of the new one, so every
+                                       run prepended another copy and the claim grew
+                                       without bound ("10099.9799.97...% HuggingFace
+                                       coverage"; 12 such spans had accumulated on
+                                       main). No prefix to preserve here.
+      "<span ...>N</span>..."          3 groups. Group 0 is leading MARKUP, which must
+                                       be preserved.
+    """
     if covered >= with_arch:
         return m.group(0)
     g = m.groups()
+    if suffix_groups == 2:
+        return _pct(covered, with_arch) + g[-1]
     return g[0] + _pct(covered, with_arch) + g[-1]
 
 
@@ -194,8 +211,13 @@ def _build_patterns(tokens, arch, covered, with_arch):
         # monster-v2 lead "317,310 arch-bearing checkpoints resolve to 552 tokens"
         (re.compile(r"(\d[\d,]*)( arch-bearing checkpoints resolve to )(\d[\d,]*)( tokens,)"),
          lambda m: _meta_pair(m, covered, tokens)),
-        # percentage claims only move when coverage drops below 100%
-        (re.compile(r"(\d+(?:\.\d+)?)%( HuggingFace coverage)"),
+        # percentage claims only move when coverage drops below 100%.
+        # [0-9][0-9.]* (not \d+(?:\.\d+)?) so the match also swallows the malformed
+        # leading runs this bug already wrote into site/*.html ("10099.9799.97...%") --
+        # with the narrow pattern the regex would match only the trailing "99.97" and
+        # leave the garbage in front of a corrected value. Greedy, but digits and dots
+        # cannot run past the '%' into markup.
+        (re.compile(r"([0-9][0-9.]*)%( HuggingFace coverage)"),
          lambda m: _pct_claim(m, covered, with_arch, 2)),
         (re.compile(r"(<span class=\"n\">)(\d+(?:\.\d+)?)(</span><span class=\"l\">checkpoints mapped</span>)"),
          lambda m: _pct_claim(m, covered, with_arch, 3)),


bong-water-water-bong:
## Audited the sibling rewriters — `_pct_claim` was the only one with this defect

Follow-up to the fix, because a one-helper bug in a file of near-identical helpers invites the
question "is this everywhere?". It is not, and the reason is a clean invariant.

The block comment directly above these helpers states the contract:

> Every replacer rebuilds the FULL match text with only the drifted number(s) swapped, and returns
> `m.group(0)` untouched when nothing drifted — so a no-change run is byte-identical (idempotent) and
> a changed run is minimal.

Every other replacer implements that as **keep-or-replace**, which cannot duplicate:

| helper | body (abridged) |
|---|---|
| `_pair` | `(g[0] if g[0] == n1 else n1) + g[1] + (g[2] if g[2] == n2 else n2) + g[3]` |
| `_frac` | `(g[0] if g[0] == fmt(covered) else fmt(covered)) + "/" + …` |
| `_of` | `(g[0] if g[0] == fmt(covered) else fmt(covered)) + g[1] + …` |
| `_num_between` | `g[0] + (g[1] if g[1] == value else value) + g[2]` |
| `_bare` | `(g[0] if g[0] == value else value) + g[1]` |
| `_meta_pair` | `(g[0] if g[0] == fmt(covered) else fmt(covered)) + g[1] + …` |
| **`_pct_claim`** (before) | **`g[0] + _pct(...) + g[-1]`** ← unconditional, no guard |

The last one is the only replacer that emits a captured group *alongside* a computed value without
comparing them first. For the 3-group caller `g[0]` is markup and that is correct; for the 2-group
caller `g[0]` is the number itself, so it was emitted twice. Hence exactly one broken helper, and
exactly the 4 files / 12 spans we see — not a broader rot.

With the guard restored, the 2-group path also satisfies the "nothing drifted → byte-identical" half:
when the digits already equal `_pct(...)`, the rebuilt string equals the original match.

## Layout regression check (both callers)

| caller | input | output | idempotent |
|---|---|---|---|
| 2-group | `100%` | `99.97% HuggingFace coverage` | yes |
| 2-group | `50%` | `99.97% …` | yes |
| 2-group | `99.97%` (already correct) | `99.97% …` (unchanged) | yes |
| 2-group | `10099.9799.97%` (corrupt) | `99.97% …` (repaired) | yes |
| 3-group | `<span class="n">100</span>…checkpoints mapped…` | `<span class="n">99.97%</span>…` — **markup preserved** | yes |
| 3-group | `<span class="n">99.97</span>…` | unchanged | yes |

The 3-group path is behaviourally untouched, which is the regression I most wanted to rule out since
both callers share the function.

bong-water-water-bong:
## This is live on the public site, in machine-readable SEO metadata

Not just a repo-hygiene issue — `https://1bit.monster/` is serving the corrupted string right now.
Fetched 2026-09-12 21:45 UTC:

```
<script type="application/ld+json">
[{"@context": "https://schema.org", "@type": "WebSite", "name": "1bit.MONSTER",
  "url": "https://1bit.monster",
  "description": "The 1-bit inference engine. One engine, any model — 568 architecture tokens,
                  2,040 HF arch strings,
                  10099.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.9799.97%
                  HuggingFace coverage, running on Ryzen AI NPUs and ROCm.",
  ...}]
</script>
```

It sits in the `WebSite` schema's `description` — the field search engines read for rich results —
not in a decorative corner of the page. The deployed `site/index.html` and `main` agree, so this is
what the crawler sees today.

Worth noting the growth direction: the span in the file is a *concatenation of one run's output per
run*, so the string gets longer every time the daily automation touches it. `site/index.html` on
`main` currently holds 48 repetitions; the live copy is the same. Whatever fixed length you see now
is not what it was yesterday and not what it would have been tomorrow.

That also means the deploy is not a second defect — the site is built from `site/*.html` on `main`,
so fixing the file on `main` fixes the live page at the next deploy. No separate change needed.

bong-water-water-bong:
## Restructured to scripts-only: the generated files cannot belong to this PR

The branch now changes **`scripts/seo_sync.py` alone**. The six `site/` repairs are dropped from the
diff, and this is a structural fix rather than a retreat — worth explaining because the earlier
description promised them.

**Why.** The SEO automation regenerates exactly those six files every ~30 minutes:

```
#2288  21:29   site/1bit-models.html, 1bit-post-1775-models.html, blog.xml, index.html, llms-full.txt, search.html
#2293  22:03   (same six)
```

So any PR holding them becomes unmergeable almost immediately. This branch hit that twice — once
against #2288 (resolved by rebasing and regenerating) and again against #2293, which had already
entered the queue. Rebasing each time just re-enters the race, and each rebase also has to re-derive the
repair on freshly regenerated output.

**Why it is still the right fix.** The repair belongs to the *generator*, and the generator is what this
commit corrects. Once the corrected rule is on `main`:

* the next `seo_sync` run rewrites `site/*.html` from it — that repair is a fixed point now, which this
  PR's verification already demonstrates (`--check`: no drift; second apply: `NO-CHANGES`);
* the next post run regenerates `site/blog.xml` and `site/llms-full.txt` from the repaired listing page,
  which is where they get the sentence from.

For the record, the fix *does* take all six files to zero malformed spans when applied — 12 in the HTML
files, plus 1 each in `blog.xml` and `llms-full.txt`. I verified that (and that `gen_rss.py` also
rewrites all 28 RSS `<pubDate>` values when regenerated, which is why those two files got a targeted
two-line repair rather than a regeneration). None of it needs to be in the diff for the correction to
take effect; it just has to be true.

**Net effect on review:** 1 file instead of 7, and a branch that can actually merge.

bong-water-water-bong:
## Verified: delegating the repair to the generators actually heals all six files

I removed the `site/` changes from this PR on the argument that the repair belongs to the generator and
the produced files catch up on their own. That is a claim about a *pipeline*, so I ran the pipeline
rather than leaving it as reasoning.

Tree with **this PR's `seo_sync.py`** plus **#2296's `gen_rss.py`**, on top of `main`:

```
before:  corrupted spans 14 (12 in site/*.html, 1 in blog.xml, 1 in llms-full.txt)
         distinct <pubDate> across 28 items: 1

  python3 scripts/seo_sync.py   # this PR    -> repairs site/*.html
  python3 scripts/gen_rss.py    # #2296      -> rebuilds site/blog.xml
  python3 scripts/gen_llms.py   # unchanged  -> rebuilds site/llms-full.txt

after:   corrupted spans 0, anywhere in site/
         distinct <pubDate> across 28 items: 7
         seo_sync.py --check: OK ... no drift
```

Each step is load-bearing and I confirmed that by removing them one at a time:

* without `gen_llms.py`, `site/llms-full.txt` still holds main's corrupted span — it is a separate
  generator from `gen_rss.py` even though both derive from the same listing page;
* `gen_rss.py` is what clears `blog.xml`, because it copies that sentence from the repaired page rather
  than from git.

So the sequence the post job already runs heals all six files once this fix is on `main` — the diff does
not need to carry generated output for the repair to happen, which was the point of reducing this PR to
its generator. The feed's dates are corrected in the same pass by #2296.

Worktree restored afterwards (`site/` back to main's content, 0 changes).

Issue #2418, opened 2026-09-15
bug(engine): the GLM-MoE-DSA gate fails when it actually runs — a single dense layer already gets top-1 wrong (15/20 vs 20/20 siblings)


bong-water-water-bong:
## Summary

`Testing/run_all.sh` has a **GLM-MoE-DSA gate that has never run in CI**, and when
it is actually given its fixture it **fails**. Minimised to a single dense layer,
the engine's top-1 does not even match the HF reference.

Found by generating the fixtures the suite skips — the e2e phases are gated on
`/tmp/onebit-*` dirs that nothing creates, so the suite reports green while these
comparators are never executed.

## What fails

```
$ python3 Testing/make_mini_glm_moe_dsa.py /tmp/onebit-glmdsa     # torch fixture
$ g++ -std=c++17 -Iinclude -Isrc -O2 Testing/cmp_glm_moe_dsa.cpp src/glm_moe_dsa.cpp \
      src/safetensors_reader.cpp src/q4nx_reader.cpp -o /tmp/cmp_glmdsa
$ echo "5 7 9 11 3" > /tmp/ids.txt
$ /tmp/cmp_glmdsa /tmp/onebit-glmdsa /tmp/ids.txt /tmp/onebit-glmdsa/logits_last.npy 20 18
[glmdsa] loaded: 4 layers, H=64, heads=4, q_lora=8, kv_lora=8, rope=4/8, v=8,
                 experts=8, top_k=2, index_topk=2, dense=1
glmdsa engine top1=171  ref top1=171  top-20 overlap=15/20 (need 18)
FAIL
```

## It is not bf16-rounding noise

The sibling mini-fixtures are built the same way (bf16 weights, HF oracle `.npy`)
and reproduce their reference **exactly**:

| lane | top-1 | top-20 overlap |
|---|---|---|
| `deepseek_v4` | 342 = 342 | **20/20** |
| `mimo_v2` | 36 = 36 | **20/20** |
| `glm_moe_dsa` | 171 = 171 | **15/20** |

So the gate's 18/20 slack is not the issue, and "bf16 vs f32 accumulation" does
not explain it — the other two lanes have zero reshuffling under identical
conditions.

## Localised to the base path, not the MoE/indexer

Regenerating with `num_hidden_layers=1` (layer 0 is the dense MLP, with the full
indexer, per the fixture's own header) makes it worse rather than better:

```
[...]

PR #2346, opened 2026-09-13
test(router): the self-check was failing on main — update it to the HRX-first policy, and wire the suite into CI

bong-water-water-bong:
### **User description**
test(router): the self-check was failing on main — update it to the HRX-first policy, and wire the suite into CI

`Testing/run_all.sh` does not pass on `main`:

```
✓ arch  ✓ discovery  ✗ router: CHECK FAILED  ✓ dtypes  ✓ sharding  ✓ rotation  ✓ iq1  ✓ tq2nz
  … 16/17 passed — 1 FAILURES
```

`Testing/router_selfcheck.cpp` fails 5 of its 20 cases, and in **all five** the shape
is the same: the router returns `[hrx_gpu, ggml_vulkan, zinc_gpu, cpu_generic]`
where the test wants the pre-HRX list. The file contains no `hrx_gpu` expectation
anywhere — it predates the policy entirely.

**The test is stale, not the router.** Every GGUF/H1B route leading with `hrx_gpu`
is deliberate and documented at the point of change: `src/model_router.cpp`'s
"HRX-first" paragraph ("HRX init() succeeds only when the HRX llama-server can
spawn AND the graph is inside the fused node set; otherwise it fails fast and the
caller cascades down the list"), the MoE-GGUF note at `:95` ("otherwise an MoE
GGUF skipped HRX entirely"), and the qwen3-GGUF lane comment above `:194`. So the
five expectations are updated to the policy — with a comment saying exactly that,
and why — rather than the policy being bent to the test.

- `openelm GGUF`, `minicpm GGUF`, `qwen3 safetensors (pilot#2 fix)` → the
  documented GGUF/H1B lane
- `MoE llama`, `MoE llama unchanged` → same lane (they still expected `hip_gpu`,
  from before the GGUF routes were unified)

Result: `ROUTER: all 20 checks passed`, and `Testing/run_all.sh` → **17/17**.

**Root cause of the rot: nothing ran the suite.** It is host-only (plain g++,
no ROCm/NPU/fixtures) and was wired into no workflow, so a routed-policy change
four weeks ago left it red with no signal. This adds a `self-checks` job to
`ci.yml` alongside `ppl-gate` — same shape, a cheap CPU-only gate that catches a
class of regression nothing else does.

Verified locally: suite 17/17, and the workflow parses (`self-checks` present with
the checked-out step).

___

### **PR Type**
Tests, Enhancement

___

### **Description**
- Update stale router self-check to HRX-first policy

[...]

PR #2421, opened 2026-09-15
fix(glmdsa): the DSA indexer rotates with the wrong RoPE convention (#2418)

bong-water-water-bong:
### **User description**
fix(glmdsa): the DSA indexer rotates with the wrong RoPE convention (#2418)

The GLM-MoE-DSA gate fails when it actually runs (`Testing/run_all.sh` skips it
unless someone generates its torch fixture): 15/20 top-20 overlap against HF, and
with a single layer the argmax is wrong outright (263 vs 56, #2418).

The cause is one convention. HF applies **interleaved** RoPE to *both* the indexer
and the main MLA attention — its own docstring says so above the call:

    Same as DeepseekV32Indexer.forward, but the indexer applies **interleaved**
    RoPE rather than the non-interleaved half-split RoPE used by DeepSeek-V3.2.

`modeling_glm_moe_dsa.py` calls `apply_rotary_pos_emb_interleave` at the indexer
(:232) and at the attention (:422). This file used `rope_interleave` for the
attention but `rope_halfsplit` for the indexer — DeepSeek-V3.2's convention, which
the engine's header had documented as "the DSA indexer convention". GLM-MoE-DSA
changed it and this lane never followed.

The indexer's rotation only affects the *scores* used to pick the top-k, so the
bug is invisible whenever nothing is dropped: at `index_topk >= seq_len` all keys
are kept and the wrong scores cannot change the result. That is exactly the
pattern in #2418 — failing at `index_topk=2`, passing at `index_topk=8`.

Verified (host-only; this lane needs no device):

| fixture | before | after |
|---|---|---|
| 4 layers, `index_topk=2` | 15/20 | **20/20** |
| 4 layers, `index_topk=8` | 20/20 | 20/20 |
| 1 layer, `index_topk=2` | **263 vs 56** ✗, 18/20 | **20/20**, 56 = 56 |
| 1 layer, `index_topk=8` | 20/20 | 20/20 |
| `Testing/run_all.sh` (fixtures present) | 25/26 | **28/28** |

Blast radius: `glm_moe_dsa_forward` has exactly two callers by text search
(`src/backend_frontier.cpp:171`, the served path, and `Testing/cmp_glm_moe_dsa.cpp:60`,
the gate). GitNexus `impact` resolved one candidate at `risk: LOW` (5 impacted, 2
direct) and returned `risk: UNKNOWN` for the other with its "absence of edges is
not evidence" note; the text search above is the confirmation the rules ask for in
that case. No signature changes.

Two things deliberately not touched: `rope_halfsplit` is now unreferenced (left in
place, no warning, and it documents the V3.2 convention the next reader may meet);
and a pre-existing `prev_topk_pos` set-but-not-used warning at `:287` — visible in
`-Wall`, unrelated to this fix, worth its own look.

Graph check: `detect-changes` cannot run against a worktree (GitNexus resolves the
single `1bit-MONSTER` repo path), so this is "unseen", not clean.

___

### **PR Type**
Bug fix

___

### **Description**
- DSA indexer rotated with wrong RoPE convention

- Switched indexer from `rope_halfsplit` to `rope_interleave`

- Matches HF GLM-MoE-DSA interleaved RoPE for indexer

- Fixes gate failures at `index_topk < seq_len`

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["DSA indexer q/k rope slice"] -- "was: rope_halfsplit (V3.2 convention)" --> B["wrong top-k scores"]
  A -- "now: rope_interleave (HF convention)" --> C["scores match HF reference"]
  C --> D["top-20 overlap 20/20"]
```

<tr>
  <td>
    <details>
      <summary><strong>glm_moe_dsa.cpp</strong><dd><code>Fix DSA indexer RoPE convention to interleaved</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/glm_moe_dsa.cpp

<ul><li>Updated header comment: both DSA indexer and main MLA attention use <br>interleaved RoPE<br> <li> Replaced <code>rope_halfsplit</code> with <code>rope_interleave</code> for indexer q and k <br>rotation<br> <li> Added long inline comment citing HF <code>modeling_glm_moe_dsa.py</code> and the <br>V3.2 convention mismatch<br> <li> Noted the bug is invisible unless the indexer drops tokens (<code>index_topk </code><br><code>< seq_len</code>)</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2421/files#diff-c6b32f815ec30d84bb3197a3c433bc8d02622c2949576ee5d2023a061d7eb29c">+19/-6</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
## Independent verification — the fix closes the gate, on the exact fixture that exposed it

I reached the same root cause independently (reading HF's `GlmMoeDsaIndexer`
against `src/glm_moe_dsa.cpp`) before finding this branch, so this is a second
confirmation of the diagnosis *and* an end-to-end check of the patch against the
HF oracle. Both are device-free.

**Comparator A/B**, same fixtures, `Testing/cmp_glm_moe_dsa.cpp` on each tree:

| fixture | before (main) | after (`57cc41c7f`) |
|---|---|---|
| 4 layers, `index_topk=2` (drops tokens) | top-20 **15/20** — FAIL | top1 171 = 171, **20/20** |
| 1 layer, `index_topk=2` | top1 **263 vs 56** — FAIL | top1 **56 = 56**, **20/20** |
| 4 layers, `index_topk=8` (drops nothing) | 20/20 | 20/20 — no regression |

The single-layer case is the sharpest: not just the tail, the **argmax was wrong
before and is exact after**. That is the case where the indexer's rotation has the
most leverage, because with `seq_len=5` and `index_topk=2` the selection decides
which keys survive at every position.

Reproduce:

```bash
python3 Testing/make_mini_glm_moe_dsa.py /tmp/onebit-glmdsa
g++ -std=c++17 -Iinclude -Isrc -O2 Testing/cmp_glm_moe_dsa.cpp src/glm_moe_dsa.cpp \
    src/safetensors_reader.cpp src/q4nx_reader.cpp -o /tmp/cmp
echo "5 7 9 11 3" > /tmp/ids.txt
/tmp/cmp /tmp/onebit-glmdsa /tmp/ids.txt /tmp/onebit-glmdsa/logits_last.npy 20 20
```

**Suite A/B** (`Testing/run_all.sh`, both trees, same `/tmp` fixtures):

| tree | result |
|---|---|
| `main` (`ccaa77a7d`) | **27/28 — 1 FAILURE** (the glmdsa gate) |
| `57cc41c7f` | **28/28 passed** |

Worth noting alongside the patch: the gate only fails when its fixture exists, and
nothing generates it — `Testing/run_all.sh` skips the phase silently otherwise and
reports green (17/17 with no fixtures, 27/28 with them). That is why this sat
unnoticed; #2346 wires the fixtureless half into CI, and the fixture-gated lanes
are still invisible to it.

The commit's own note — "invisible until the indexer drops tokens
(`index_topk < seq_len`)" — is exactly what the A/B shows, and it matches the
V3.2-vs-GLM-MoE-DSA distinction HF spells out above
`apply_rotary_pos_emb_interleave`.

PR #2433, opened 2026-09-16
fix(bench): the oracle harness scored a truncated window AND could silently score a stale prompt

bong-water-water-bong:
## What

Both oracle accuracy harnesses scored the **first N characters** of each arm's output rather
than the output: `oracle_accuracy_model.sh` at **200** chars, `oracle_accuracy_0_6b.sh` at **90**.
Every model in the set emits a `<think>` reasoning block before its answer, so the scored window
usually held only the assistant's preamble — and the answer arrived past the cut.

This is a **measurement-validity** defect. It is not an engine result.

## Evidence

**1. The cap bound on every row.** Re-analysing the committed 2026-09-16 TSVs, counting rows where
either stored arm text reaches the 200-char cap:

| model | rows | rows hitting the cap | both arms scored N |
|---|---:|---:|---:|
| Qwen3-4B | 20 | **20** (flm 20, native 20) | 8 |
| Qwen3-8B | 20 | **20** (flm 19, native 20) | 9 |

**2. The decisive case, re-scored from the raw that was still on disk** — same bytes, same token
ids, only the window changed:

```
Qwen3-4B, "The capital of Germany is"
  full decoded text          : 571 chars, expected "berlin"
  OLD  (first 200 chars)     : N     <- what the 2026-09-16 run recorded
  NEW  (full text)           : Y
  discarded text contains    : "Berlin is the capital of Germany" at ~char 250
```

The model was right and the harness recorded it wrong. The same mechanism produces individual
cross-arm disagreements: on "The capital of Japan is", FLM's reasoning lands *"Japan's capital is
Tokyo"* inside the window (Y) while native's is still winding up (N) — the arms differ in how fast
they reach the point, not in whether they know it.

## What this does *not* establish

**Native does not thereby meet FLM on accuracy.** It establishes that the recorded
**4B `8/20` vs `11/20`** and **8B `6/20` vs `8/20`** are **not measurements of accuracy**, so
neither the deficit nor parity can be claimed from them. The true numbers are unknown until
re-measured.

**"Native is behind FLM on 4B/8B accuracy" should be read as UNVERIFIED, not refuted.** The
truncation was symmetric, so the ordering it produced is still a real signal — about early
emission, not correctness. This matters because that gap is currently ranked as a parity
blocker; it should not be, on this evidence.

## Change

- Cap is now `ORACLE_TEXT_CHARS` (default **4000**, `0` = unlimited) instead of hard-coded 200/90.
  `ORACLE_TEXT_CHARS=200` reproduces the old behaviour for an A/B.
- New per-arm verdict scoring the **answer region** — text after the last `</think>`, else the whole
  text — reported *alongside* the existing whole-text verdict. That is the stricter "did it actually
  answer" signal the cap had been silently proxying for.
- New TSV columns `flm_ans`/`native_ans` (`rl_ans`/`dense_ans` for the 0.6B harness). **Existing
  columns and summary lines keep their names and positions**, so the older series and any reader of
  it remain comparable.

## Verification

Done: `bash -n` on both scripts; TSV header/format arity check (**10/10/10** and **14/14/14**); the
direct re-score of the real Germany row above.

**Not done — stated plainly: neither harness has been re-run on the device.** `/dev/accel/accel0` is
shared (rule 4) and the dense lane was live while this was written, so I did not contend for it. The
re-run is the next step, not a claim made here.

Also NOT settled: the **0.6B `20/20`** was measured under its 90-char cap. Whether it survives is
untested and it should not be quoted as settled in the meantime. I flagged it rather than leaving the
sibling harness silently inconsistent, but the fix to it is the same three lines.

Finding recorded in `benchmarks/RESULTS-oracle-scorer-truncation-2026-09-16.md`.

Base is `goal/runlist-decode-wire` because neither harness exists on `main` — both are introduced by
this campaign branch.

---

## SECOND DEFECT (added after the first was verified on device): a STALE prompt could be scored silently

While re-measuring on device I hit a worse failure than the truncation, in the same harness.

The harness wrote its ids to a **fixed** path (`/tmp/oam_ids.txt`) and tested it only for
**non-empty**. When the tokenize step failed, the previous run's file was still there, passed the
guard, and **the engine was scored on a completely different question** — recorded as an ordinary
row, with nothing in the TSV to say so.

```
11:15  /tmp/oam_ids.txt written  -> "The capital of Germany is"
11:52  a run asks "2 + 2 =" and "The capital of Italy is"
       /tmp/oam_ids.txt unchanged (mtime 11:15:59)
       every row returns the SAME answer: "...the capital of Germany..."
```

The engine is **not** at fault, and that mattered enough to prove rather than assume: handed its
own unique ids file it answers each prompt differently, and France vs `2 + 2 =` diverge at token 5
of the decode. The harness was feeding it one prompt over and over.

The silent tokenize failure underneath was `UndefinedError: 'user_system_prompt' is undefined` —
Qwen3-4B and 1.7B reference an optional template variable. **The weak guard is what turned a loud
error into a wrong measurement.**

Second consequence of the same fixed paths: **two concurrent runs clobber each other** — two
models, or another agent on this box.

### What changed

- **Per-invocation scratch**: `WORK=$(mktemp -d)`, with `IDS`/`TPL`/`NAT_RAW`/`FLM_RAW` inside it and
  an `EXIT` trap. Concurrent runs can no longer collide.
- **The ids file is removed before the tokenize**, so the guard asks *"was THIS prompt tokenized"*
  rather than *"is there a file"*. A failed tokenize now reports `TOKENIZE EMPTY` and writes an ERR
  row; it cannot fall back to a stale prompt.
- The chat-template call supplies `user_system_prompt=""` / `tools=None` with a ChatML fallback.
  Without it the call **raises** for Qwen3-4B/1.7B — exactly the failure the guard was hiding.
  *Overlaps with the goal lane's uncommitted I1 change to this file; whichever lands first, the
  other's hunk is a no-op.*

### Verified on device, after the fix

Same model, same engine, same prompts:

| prompt | before | after |
|---|---|---|
| "The capital of **France** is" | German answer → **N** | reasons about France, *"Paris is the capital"* → **Y** |
| "The capital of **Italy** is" | **identical** German answer → **N** | reasons about Italy, *"Rome is the capital"* → **Y** |

Prompt-specific, correct, and credited by the answer-region rule.

**This does not by itself establish the native arm's true accuracy** — it removes two ways the
recorded number could be wrong. It does mean the recorded **4B 8/20 and 8B 6/20 are not trustworthy
as measurements.** The full 20-prompt re-measure is running.

---

## RE-MEASURED ON DEVICE — gap #2 is withdrawn

Both models re-run with this harness, same prompts, 128-token budget, greedy, `NPU_RUNLIST=1`:

| model | arm | recorded | re-measured |
|---|---|---|---|
| Qwen3-4B | FLM oracle self-check | 11/20 | **20/20** |
| Qwen3-4B | native | 8/20 | **20/20** |
| Qwen3-8B | FLM oracle self-check | 8/20 | **20/20** |
| Qwen3-8B | native | 6/20 | **18/20** |

**The scorecard's gap #2 — "Qwen3-4B / 8B answer accuracy below FLM" — is withdrawn.** Both arms
were under-measured by the same two defects; the recorded *delta* was never a capability
difference.

**8B's two misses are budget, not knowledge.** The 128-token window closes while the model is still
reasoning, so the answer never appears:

- `3 + 4 =` (want `7`) — *"…I should check if there's any context or hidden meaning…"*
- `The opposite of black is` (want `white`) — *"…red and green are opposites… However, the user…"*

Neither is a wrong answer; FLM just reaches the answer sooner. That makes the score sensitive to
`ntok` in a way the earlier record did not say — worth knowing for anyone scoring reasoning models
this way.

**Not re-run: Qwen3-1.7B.** Its recorded 10-vs-9 predates these fixes and carries the same two
defects, so it is not comparable to the 4B/8B numbers and should not be cited either way until
re-run.

Per-row TSVs and RESULTS docs:
`benchmarks/RESULTS-oracle-4b-accuracy-retraction-2026-09-16.md`,
`benchmarks/RESULTS-oracle-8b-accuracy-retraction-2026-09-16.md`.

___

### **PR Type**
Bug fix, Documentation, Tests

___

### **Description**
- Fixes oracle harnesses that scored truncated output windows

- Per-invocation scratch dirs remove stale/concurrent prompt clobbering

- Adds answer-region verdict alongside whole-text accuracy verdict

- Fixes tokenizer eating special tokens after BPE spans

- Documents retracted 4B/8B accuracy gaps and dense-arm failure modes

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["oracle_accuracy_model.sh / _0_6b.sh"] -- "ORACLE_TEXT_CHARS cap" --> B["Full-text scoring"]
  A -- "mktemp scratch + rm ids" --> C["No stale or clobbered prompts"]
  B -- "ans_region after last think" --> D["New *_ans TSV columns"]
  E["tokenize.cpp</br>BPE span eats special"] -- "cap match at next special" --> F["<|im_end|> preserved"]
  G["RESULTS-*-2026-09-16.md"] -- "evidence and retractions" --> H["4B 20/20, 8B 18/20 retraction"]
  I["npu_engine_universal.cpp"] -- "NPU_DUMP_LOGITS_PERSTEP" --> J["Per-step top-8 + margin"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>npu_engine_universal.cpp</strong><dd><code>Adds per-step logits dump with top-8 margin</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-63677a636d5ff19712cf034cc1e2c15a96b194137caee54d8bd6883b0b76a10b">+26/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Bug fix</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>tokenize.cpp</strong><dd><code>Caps BPE match before upcoming special token</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-bb136ab471838467f8d177aeacb49f96372963b08321bdef4f098f4fa18e440e">+17/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>oracle_accuracy_0_6b.sh</strong><dd><code>Configurable text cap, scratch dirs, answer-region verdict</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-b6d20346d40df96f6a7a479dd628f5f3898161a3603f9bda5ccf34e912780c7d">+79/-28</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>oracle_accuracy_model.sh</strong><dd><code>Removes stale ids, configurable cap, answer-region verdict</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-419aab9d0e8322fb5229489d5fcffd540d9b874d9b2dc8944e410119ada38581">+99/-25</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>8 files</summary><table>
<tr>
  <td><strong>RESULTS-06b-arms-diverge-step8-2026-09-16.md</strong><dd><code>Records arms agreeing eight tokens before divergence</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-89e898855bb0669810b7668072016e4842b65114cc0682234665664e9aad1ca0">+46/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-06b-divergence-is-a-near-tie-2026-09-16.md</strong><dd><code>Measures near-tie fork, refutes scale hypothesis</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-2e90c6c7b984cb3ef0b52a2786bd148ff450dd105e871f1f10a7d85bff3b0823">+220/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-8b-budget-attribution-VERIFIED-2026-09-16.md</strong><dd><code>Verifies 8B misses are budget exhaustion</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-b1c79718a67f19a1740f3caed6c28b0f6cc93d2ccfde536b7a66181d6be82e8d">+52/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-gdn-bf16-cannot-explain-the-NaN-2026-09-16.md</strong><dd><code>Refutes bf16 GDN recurrence NaN root cause</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-98dd6b1c5beb4309aa11e23871ea6df2fb14b8cecdb1555bbe7bbf2b861bce4b">+114/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-oracle-4b-accuracy-retraction-2026-09-16.md</strong><dd><code>Retracts 4B accuracy gap as harness artifact</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-ffa48366d81f1b1c2d044983e5088cd983043ec9e37d9bacd5af0c61d3445c29">+68/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-oracle-8b-accuracy-retraction-2026-09-16.md</strong><dd><code>Retracts 8B accuracy gap as harness artifact</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-f43a76720092e2fd573b9accad4948c2e1491697449511da23e852679750470e">+61/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-oracle-scorer-truncation-2026-09-16.md</strong><dd><code>Documents truncated scoring window harness defect</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-8f4f0c6ebe2c54b6f1d55f0dc8812d6fadac84bcfb7f74eebbb5e64410e9643e">+91/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-tokenizer-special-span-2026-09-16.md</strong><dd><code>Documents special-token span tokenizer corruption</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2433/files#diff-c9a4a4da60ad050b1bf7a892e23a9901b9e6097f1709cb1cb06fe4c24e5ac1d5">+199/-0</a>&nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Posting the reason this PR has been unable to merge for two days, since nothing in the UI says it.

**No `CI` and no `CodeQL` have ever run on this branch.** Its workflow history is `PR Agent` and
`Scope Guard` only (plus one `NPU Benchmark`); the head carries exactly two check-runs, and
`Scope Guard` is the only one of the four required names that is present. So the PR cannot satisfy
the ruleset, and `autoMergeRequest` is NULL.

The cause is a base-branch filter plus a retarget: `ci.yml` and `codeql.yml` trigger on
`pull_request: branches: [main]` — the *base* — with the default event types, and retargeting a PR
to `main` is an `edited` event that neither subscribes to. `Scope Guard` and `PR Agent` have no
branch filter, which is why they ran and the PR looks tested.

**The fix on this branch:** any push re-triggers all of it via `synchronize` — and since
`mergeStateStatus` is `DIRTY`, a rebase onto `main` is needed anyway. Reopening the PR would also
work. `auto-enqueue.yml` already handles the retarget for its own purposes (it subscribes to
`edited` since #2536), so once CI runs, the enqueue path takes care of itself.

Filed generally as #2564 — I am not proposing a change to `ci.yml` there, because subscribing it
to `edited` would re-run the full `C++` job on every title or body edit.

bong-water-water-bong:
Two things measured today, because they change what the remedy for this PR is.

**1. The defect this PR fixes is still live on `main`.**

`benchmarks/oracle_accuracy_model.sh` on `64939fbc5`/`1cd7fdd61` still scores the first 200 characters — `[:200]` at line 82 and `cut -c1-200` at line 92 — and `benchmarks/oracle_accuracy_0_6b.sh` still uses `cut -c1-90` at lines 89/100/107. Neither file mentions `ORACLE_TEXT_CHARS`. So this is not a stale PR against already-fixed code: the measurement-validity defect is in `main` right now, and the fix commit `c28b2fda2` is the only place it exists.

**2. This branch cannot merge, at all — so "push to re-trigger CI" is not the remedy here.**

| | |
| --- | --- |
| `changed_files` | **17,358** |
| commits ahead of `main` | 1,264 (307 behind) |
| merge-base with `main` | `1e48a85b6`, 2026-09-09 — nine days old |
| `mergeable` | `false`, `mergeStateStatus: DIRTY` |

The merge base predates the squash-merges that landed this branch's own base, so `main` and this branch share no usable ancestry: the diff is the whole tree, not the fix. A rebase would replay 1,264 commits that are mostly already in `main` under different SHAs. My earlier comment on this PR said "any push re-triggers all of it via `synchronize`" — that is still true mechanically, and it is still the answer for the *CI* gap, but it does nothing for the merge, and I should have said so.

**What would work.** A fresh branch off current `main` carrying only the fix: `benchmarks/oracle_accuracy_*.sh` plus the results doc. From `c28b2fda2` that is 2 script edits (the cap becomes `ORACLE_TEXT_CHARS`, default 4000, `0` = unlimited; a new per-arm answer-region verdict; `flm_ans`/`native_ans` TSV columns) and a new RESULTS file — small and reviewable, against 17,358 files here. `ORACLE_TEXT_CHARS=200` reproduces the old behaviour, so the re-land keeps its A/B.

I have not created that branch: the commits are authored by `pi agent 07e844`, so this lane's fix is this lane's to re-land, and duplicating it from my side would be a second PR for the same change. Flagging it because the alternative — this PR eventually merging — is not available.

bong-water-water-bong:
**Correction to my earlier comment here.** I wrote that "any push re-triggers all of it via `synchronize`". For this PR that is wrong, and the distinction matters because it is the state this PR is in.

GitHub documents it directly: *"If the pull request has a merge conflict, workflows will not run on `pull_request` activity. The merge conflict must be resolved first."* ([events that trigger workflows](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request)). This PR reads `mergeable=false` / `mergeable_state=dirty`, so **no** `pull_request` workflow runs on it right now — not CI, not Scope Guard, not PR Agent — and a push while it conflicts starts nothing. The check list simply shows the required contexts as absent, which is why it looks like "waiting" rather than "blocked".

So the order is: resolve the conflict (rebase or merge `main`) **first**; the push that comes with it is what then re-runs everything. I also overstated a related point on #2564 — that the silence after this morning's automatic retarget proved no event was delivered. Every workflow's type list excludes `edited`, and a conflicted PR runs no `pull_request` workflow at all, so the silence cannot distinguish the two. The corrected analysis, with the two traps kept separate, is in [that comment](https://github.com/1bit-MONSTER/1bit-MONSTER/issues/2564#issuecomment-5725660893).

PR #2434, opened 2026-09-16
fix(npu): MoE harness packed layer 0 and ran layer 1 ELF; + the conv1d step is never applied

bong-water-water-bong:
## What

`MoERuntimeLayerEngine::init` called all four weight packers with **layer 0**, then warmed
`moe_layer_ctx1.elf`; `moe_smoke` called `forward(1)`. The harness ran **layer 1's program with
layer 0's weights**. Nothing errored — the two layers share a layout, so it produced plausible
wrong output instead of a failure.

## The convention that hid it

The parameter was named `ctx_len` and documented as "the 1-based context length". For the **dense**
engine that is true: `RuntimeLayerEngine` loads `layer_ctx<N>.elf` where N really is a context
length (`layer_ctx444.elf`, `layer_ctx2086.elf`, …). The MoE engine reuses that filename shape with a
different meaning, and the code believed the wrong one:

```
$ ls npu-infer/captures/txn-elfs-moe35b/moe_layer_ctx*.elf | wc -l   ->  40
$ indices present                                                    ->  0 1 2 ... 38 39
$ model config num_hidden_layers                                     ->  40
```

Forty files, indices 0..39, forty layers. **`moe_layer_ctx<N>.elf` is layer N.** The header's own doc
comment said "single layer, **linear layer 0**" — matching the packed weights and contradicting the
ELF actually loaded. The two halves of the harness disagreed and nothing could notice.

**Why it ran instead of failing:** `moe_layer_ctx0.elf` and `moe_layer_ctx1.elf` are both 103,104
bytes — identical size, because both layers are `linear_attention` (addendum 148). A wrong-layer run
yields a *wrong answer*, never a *loud failure*. Same silent-failure class as the two attention guards
this lane already added.

## Change

- `ctx_len` → `layer` throughout the MoE engine, with the two conventions documented in the header —
  the old name is what allowed the conflation.
- `init(..., int layer = 1)` selects the packed layer and passes **that same value** to
  `ensure_layer_kernel`, so the ELF warmed at init matches the packed weights.
- The engine records `packed_layer_` and **refuses** any other layer's ELF, naming both layers and the
  fix. The weight/router/norms BOs hold one layer, so this is never legitimate under the current
  single-BO design.
- `moe_smoke` takes the layer from `argv[5]` or `MOE_LAYER` (default **1**) and uses one value for both
  packing and forward, printing it.

**Default 1 is deliberate: it is the single-variable change.** The harness already loaded layer 1's
ELF; only the weights are now made to match it. Defaulting to 0 would have changed which layer is
exercised as well.

## Verification

**Done:** `g++ -std=c++17 -fsyntax-only -Wall -Wextra` on both TUs (**rc=0**, no new warnings); full
compile **and link** against the runlist XRT 2.26.0 set (**rc=0**, binary produced); the ELF premise
confirmed from the artifacts (40 files vs 40 layers).

Blast radius by text search (the index returns `risk: LOW` with 1 caller for the MoE method, and an
`UNKNOWN` for the **dense** class's same-named method): the MoE `ensure_layer_kernel` has exactly two
callers — `init` and `forward`, both updated, same class. The dense `RuntimeLayerEngine` is untouched.
`MoERuntimeLayerEngine` has exactly one client, `moe_smoke`, updated here. `init` gained a *defaulted*
parameter, so it stays source-compatible.

**Not done — stated plainly: the binary has not been run.** `/dev/accel/accel0` is held by the dense
lane (rule 4), so the refusal path and the effect of a now-consistent layer on the logits are
**unexercised**. What is verified is that it compiles, links, and that the layer it packs is the layer
it loads.

## Scope note

This is **not** the cause of the MoE NaN — addendum 148 already said so and this does not contradict
it. It is the reason any comparison against a per-layer reference could not have been valid: the
reference was for one layer and the harness computed with another layer's weights. Such a comparison
could not have passed, and its failure was not evidence about the ELF, the kernels, or the device.

Incidental: `moe_smoke.cpp`'s build recipe compiles `model.c` with `g++`, which forces C++ and fails on
its `void*` conversions. `gcc` for `model.c` and `g++` for the rest links cleanly.

Recorded in `benchmarks/RESULTS-moe-layer-pack-mismatch-2026-09-16.md`.

Base is `goal/runlist-decode-wire` because these files are introduced by that campaign branch.

---

## SECOND COMMIT — a finding, not a fix: the MoE path never applies the linear-attention conv1d

While in this code I answered addendum 148's open question ("where the conv1d/SSM weights ARE read
from"). **Answer: nowhere — not on the device, and not on the host either.**

Found by comparing against the **working reference** rather than decoding the vendor ELF further.
The dense GDN path (`gdn_attn_step`, Qwen3.5-4B, whose fallback prefill boots correctly) applies
the causal depthwise conv1d **in plain C++ on the host**, from `gdn_conv_w[l]` loaded as host floats
straight off the model's `ssm_conv1d` tensor. **Conv is a host op in the reference implementation** —
which is exactly why the layer ELF has no read for it.

The MoE path has none: `runtime_layer_moe.cpp`'s single `conv` mention is a comment, and
`moe_smoke.cpp` and the header have zero. `forward()` packs the BOs and submits the ELF; nothing
applies conv.

It cannot be hiding inside the ELF. The `.txn` is **98,544 B of the ELF's 103,104 B**, present
verbatim (verified by substring search), leaving only **4,560 B** of non-TXN space — against the
**65,536 B** the conv1d weights need. I also re-derived the read set with the decoder: across all
630 patches, `arg3 @0` appears only as an **S2MM (write)** of 16,512 words, never as an MM2S (read),
and no patch anywhere reads a 65,536-byte region.

> **Correction (same day).** The first version of this argument was *"65,536 B = 32,768 words >
> the TXN's 24,636 words, so it is arithmetically unable to fit."* **That was wrong and is
> removed.** The TXN's words are **uint32** (98,544 / 24,636 = exactly 4.000 B/word), so 65,536 B is
> **16,384 words — fewer than 24,636**, and it could have fit on that count. The conclusion survives
> on the ELF byte budget, which is sound and verified. Recording the bad reasoning rather than
> quietly swapping it in.

So the packer's 65,536-byte prefix is not merely dead weight in the wrong half of the BO: **it is
weights for a step that never runs**, written into the region the ELF uses as S2MM scratch. The
clobber is harmless *precisely because* nothing reads it.

This is a **missing computation**, not a binding error — which fits the observation that the arg
binding decoded clean and consistent while the output is wrong.

**Not established:** that this is the only defect, or that it is specifically the NaN source. The
wrong-layer ELF/weights mismatch fixed in the first commit confounds any conclusion drawn from the
harness's output before that fix.

**Next step, and it is cheap:** apply the same host conv1d the dense path already implements, using
the `ssm_conv1d` the packer already loads, and re-run `moe_smoke` — then the comparison is between
two paths that at least compute the same function.

Recorded in `benchmarks/RESULTS-moe-conv1d-never-applied-2026-09-16.md`.

___

### **PR Type**
Bug fix, Documentation

___

### **Description**
- MoE engine now packs and runs the same layer

- `ctx_len` renamed to `layer`; mismatch refused loudly

- Add host `logits_host` for 3-D Q8_0 lm_head

- Region-B base and weight BO size env-overridable

- Extend probes: zero-weight/norms, act BO scan

- Record nine benchmark notes of measurements and retractions

___

### Diagram Walkthrough

```mermaid
flowchart LR
  SMOKE["moe_smoke: layer from argv[5] / MOE_LAYER"] -- "passes layer" --> INIT["init(..., int layer=1)"]
  INIT -- "packers use layer" --> PACK["weight/router/norms BO: one layer"]
  INIT -- "records" --> PL["packed_layer_"]
  FWD["forward(layer)"] -- "ensure_layer_kernel" --> GUARD{"layer == packed_layer_?"}
  GUARD -- "no" --> REFUSE["REFUSING layer N: report and return false"]
  GUARD -- "yes" --> RUN["runlist: MLIR_AIE layer ELF"]
  RUN -- "3-D lm_head" --> HOST["logits_host: stream Q8_0 dequant"]
  HOST -- "logits" --> SCAN["moe_smoke: NaN / all-zero / act page scan"]
  ENV["MOE_REGION_B_BASE, MOE_WEIGHT_BO_BYTES, MOE_ZERO_WEIGHTS"] -- "A/B probes" --> INIT
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>runtime_layer_moe.cpp</strong><dd><code>Rename ctx_len to layer, refuse ELF mismatch, add host lm_head</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-d551dba96ae100adf06007e85faf51aef80190abe205e49122ca1a675cba9212">+191/-26</a></td>

</tr>

<tr>
  <td><strong>moe_smoke.cpp</strong><dd><code>Take layer from argv/MOE_LAYER, prefer host lm_head, scan act BO</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-585e1f3f2e69a46f3a98d50826c55e638ec19d09d1aa210bed4d037389c2c1ca">+101/-14</a></td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>9 files</summary><table>
<tr>
  <td><strong>runtime_layer_moe.h</strong><dd><code>Document layer-index convention, new signatures, packed_layer accessor</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-d8611db938d9dfce0277728eb28390b01b2968490048e663b692664a6592bfbe">+37/-5</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe-layer-pack-mismatch-2026-09-16.md</strong><dd><code>Note the harness packed layer 0 and ran layer 1</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-d7af102d7d1d44a86d7547562764898c8b45153208654389af6bd4df1faf887b">+105/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-lmhead-ndim-silent-skip-2026-09-16.md</strong><dd><code>Note 3-D lm_head silently skipped by ndim == 2 guards</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-2e2bc805e3e951fea307279dfb105a3c21377dda76ad7499753bc0aaaec4e5ae">+307/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe-bo-nan-survey-2026-09-16.md</strong><dd><code>Survey every BO for NaN; state is f32, 98.2% NaN</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-f8c5956c76a766077fc787c8b77afcfc048fbd06be68cbfaa2e9565d4b5af2ea">+121/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe35b-flm-real-arg-bindings-2026-09-16.md</strong><dd><code>Capture FLM's real arg bindings; refute BO-size hypothesis</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-05be2b89a80692fc491b03a3ada6c6f9dc4354d88816522491e9229742468912">+101/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe-conv1d-never-applied-2026-09-16.md</strong><dd><code>Note the linear-attention conv1d step is never applied</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-50fc101348a2f4189ff6e4ead8744a6fa2260e3cd2ee1c3f43e70a1599fa9cde">+99/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe35b-argmap-evidence-2026-09-16.md</strong><dd><code>Arg-map evidence and the unresolved patch-table contradiction</code></dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-4f47b7ad51c9bcc936ae7a9190636967681c94e4bc53ed7b956c9323554195e7">+81/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe35b-elf-does-not-read-weights-2026-09-16.md</strong><dd><code>Note the layer ELF never consumes the weight BO</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-ee90970c3b4f40c676a0d6545d8dfe57bbdeb653ebc2f68d4068168e5cd1bf5c">+91/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe35b-nan-root-cause-arg-order-2026-09-16.md</strong><dd><code>Claim arg order was the NaN root cause, later superseded</code>&nbsp; </dd></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-17b27854ae29a495319b1a9412a3ee908ca6c277750a1d2e07ed3d7d3aeaf9b1">+67/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Additional files</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>RESULTS-both-moe-elfs-NaN-on-all-zero-BOs-2026-09-16.md</strong></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-18f5c7ded03e06f98d24e913d0b366ee34c30f47705e053b24af7eb1163040ec">+71/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe-elf-NaNs-on-all-zero-input-2026-09-16.md</strong></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-cd1b0b45fd42365b86debe3bbcd20d5caa7ca5b757c3c7fa39cea2ace88fc9a8">+67/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>RESULTS-moe-nan-is-not-in-the-data-path-2026-09-16.md</strong></td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2434/files#diff-154791e6dfaa8fd854fb5b8694714feb949aca3ef1c7093a8c9b7b50d19c2774">+70/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Posting the reason this PR has been unable to merge for two days, since nothing in the UI says it.

**No `CI` and no `CodeQL` have ever run on this branch.** Its workflow history is `PR Agent` and
`Scope Guard` only (plus one `NPU Benchmark`); the head carries exactly two check-runs, and
`Scope Guard` is the only one of the four required names that is present. So the PR cannot satisfy
the ruleset, and `autoMergeRequest` is NULL.

The cause is a base-branch filter plus a retarget: `ci.yml` and `codeql.yml` trigger on
`pull_request: branches: [main]` — the *base* — with the default event types, and retargeting a PR
to `main` is an `edited` event that neither subscribes to. `Scope Guard` and `PR Agent` have no
branch filter, which is why they ran and the PR looks tested.

**The fix on this branch:** any push re-triggers all of it via `synchronize` — and since
`mergeStateStatus` is `DIRTY`, a rebase onto `main` is needed anyway. Reopening the PR would also
work. `auto-enqueue.yml` already handles the retarget for its own purposes (it subscribes to
`edited` since #2536), so once CI runs, the enqueue path takes care of itself.

Filed generally as #2564 — I am not proposing a change to `ci.yml` there, because subscribing it
to `edited` would re-run the full `C++` job on every title or body edit.

bong-water-water-bong:
Two things measured today, because they change what the remedy for this PR is.

**1. The defect this PR fixes is still live on `main`.**

`npu-infer/src/runtime_layer_moe.cpp` on `main` still carries the broken binding, unchanged:

    // Round-73 MoE arg order: slot3=weight, slot4=act, slot5=router,
    // slot6=norms, slot7=kv/state.
    run.set_arg(3, (const xrt::bo&)*bo_weight_);
    run.set_arg(4, (const xrt::bo&)*bo_act_);

No line in `main`'s copy of that file mentions the root cause, the arg-order swap or `MOE_ARG_LEGACY_ORDER`. So the all-NaN 35B layer is not fixed in `main`; the fix commit `1ae82a6e1` is the only place it exists.

**2. This branch cannot merge, at all — so "push to re-trigger CI" is not the remedy here.**

| | |
| --- | --- |
| `changed_files` | **17,357** |
| commits ahead of `main` | 1,294 (307 behind) |
| merge-base with `main` | `1e48a85b6`, 2026-09-09 — nine days old |
| `mergeable` | `false`, `mergeStateStatus: DIRTY` |

The merge base predates the squash-merges that landed this branch's own base, so `main` and this branch share no usable ancestry; the diff is the whole tree, not the fix. My earlier comment here said "any push re-triggers all of it via `synchronize`" — mechanically true and still the answer for the CI gap, but it does nothing for the merge, and I should have said so.

**What would work.** A fresh branch off current `main` carrying only the fix: the two `set_arg` calls swapped in `npu-infer/src/runtime_layer_moe.cpp` plus the root-cause comment, the `moe_smoke.cpp` changes (+61), and the results doc — against 17,357 files here. Carry the commit's own limits into the re-land: it explicitly does not establish numerical correctness (single-layer argmax 193722 vs the stored reference 76740, and that reference is itself suspect because it was derived under the broken order), and args 5/6/7 were never re-derived against FLM's order. Keeping `MOE_ARG_LEGACY_ORDER=1` preserves the two-way falsifiability.

I have not created that branch: the commits are authored by `pi agent 07e844`, so this lane's fix is this lane's to re-land, and duplicating it would be a second PR for the same change. Flagging it because the alternative — this PR eventually merging — is not available.

bong-water-water-bong:
**Correction to my earlier comment here.** I wrote that "any push re-triggers all of it via `synchronize`". For this PR that is wrong, and the distinction matters because it is the state this PR is in.

GitHub documents it directly: *"If the pull request has a merge conflict, workflows will not run on `pull_request` activity. The merge conflict must be resolved first."* ([events that trigger workflows](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request)). This PR reads `mergeable=false` / `mergeable_state=dirty`, so **no** `pull_request` workflow runs on it right now — not CI, not Scope Guard, not PR Agent — and a push while it conflicts starts nothing. The check list simply shows the required contexts as absent, which is why it looks like "waiting" rather than "blocked".

So the order is: resolve the conflict (rebase or merge `main`) **first**; the push that comes with it is what then re-runs everything. I also overstated a related point on #2564 — that the silence after this morning's automatic retarget proved no event was delivered. Every workflow's type list excludes `edited`, and a conflicted PR runs no `pull_request` workflow at all, so the silence cannot distinguish the two. The corrected analysis, with the two traps kept separate, is in [that comment](https://github.com/1bit-MONSTER/1bit-MONSTER/issues/2564#issuecomment-5725660893).

PR #2439, opened 2026-09-17
fix(npu): create the build dir before its first use, so a fresh clone can build

bong-water-water-bong:
## The bug

`engine/npu/build_npu.sh` sets `BUILDDIR` at the top and uses it immediately — the three "one-time"
compiles write `$BUILDDIR/{dequant_q4nx,gemm_npu_instructions,zaya_decode}.o` — but the only
`mkdir -p "$BUILDDIR"` sat **below** them, just before the model loop. With `set -e` and no directory,
a tree that has never been built dies on the very first `gcc`:

```
$ bash engine/npu/build_npu.sh
gcc -c -O3 -o .../engine/npu/build/dequant_q4nx.o .../dequant_q4nx.cpp
Assembler messages:
Fatal error: can't create .../engine/npu/build/dequant_q4nx.o: No such file or directory
rc=1
```

Reproduced on a fresh worktree of `380f0b90f`. `engine/npu/build` is gitignored by the general
`build/` rule (`.gitignore:16`), so a fresh clone genuinely lacks it. The script has only ever worked
because the directory happened to already exist on this box.

> **Correction (2026-09-17).** An earlier version of this body, and the commit message, said the
> directory is "not gitignored". That is wrong — it is. My check ran while the directory did not
> exist, and `build/` is a *directory-only* pattern, so git could not match it and reported nothing.
> This is the same shape as the bug being fixed: a check whose silence meant "the directory is not
> there", not "the rule does not apply". The conclusion is unchanged — a fresh clone lacks the
> directory either way — but the stated mechanism was false.

## The fix

`mkdir -p "$BUILDDIR"` moved above the one-time compiles; the redundant one before the model loop is
removed. Two-line change.

## Verified on the same never-built worktree

```
$ rm -rf engine/npu/build && bash engine/npu/build_npu.sh
BUILD_RC=0
--- errors ---        (none)
--- binaries ---      19
```

## Why it survived, which is the more useful finding

**No CI job invokes `build_npu.sh`.** The only workflow whose text mentions it is
`xclbin-provenance.yml`, and that is a comment about `build_npu_ternary.sh` (a different script). So
nothing builds the NPU engine on a clean tree, and build rot here is invisible until someone clones —
which is exactly the shape of #2382 (the installer naming four CMake targets that no longer existed:
ready-made for anyone who had not cloned recently).

This is found work, not my own: it is item 2 of `NOTICE-pr2435-conflicts-resolved-2026-09-16.txt` in
the coordination mailbox, which flagged it as out of scope for that merge and left it for "anyone who
wants fresh checkouts to build".

## Not changed

`docs/wiki/npu-architecture.md` still says `build_npu.sh # All 5 variants` where the script builds 18
(+ a default). I left it: that file's own header declares it a historical research snapshot, "not the
current production architecture", last updated 2026-07-24. Chasing a count in a self-declared archive
is noise; `docs/journey.md` already says 19. Flagging it here so the choice is visible rather than an
oversight.

___

### **PR Type**
Bug fix

___

### **Description**
- Move `mkdir -p "$BUILDDIR"` above the one-time compiles

- Fresh clones no longer fail on the first `gcc`

- Remove redundant `mkdir` before the model loop

- Comment records the `set -e` failure and why it survived

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["build_npu.sh"] -- "mkdir -p $BUILDDIR early" --> B["engine/npu/build created"]
  B -- "one-time compiles" --> C["dequant_q4nx.o / gemm_npu_instructions.o / zaya_decode.o"]
  C -- "model loop" --> D["19 npu_engine_* binaries"]
  A -- "removed duplicate mkdir" --> E["no redundant call"]
```

<tr>
  <td>
    <details>
      <summary><strong>build_npu.sh</strong><dd><code>Create NPU build dir before first compile</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

engine/npu/build_npu.sh

<ul><li>Adds <code>mkdir -p "$BUILDDIR"</code> right after <code>REPO_ROOT</code> is resolved, before <br>the one-time compiles that first write into <code>$BUILDDIR</code>.<br> <li> Removes the now-redundant <code>mkdir -p "$BUILDDIR"</code> that sat just before <br>the model loop.<br> <li> Adds a comment explaining the <code>set -e</code> + missing-directory failure <br>(<code>Fatal error: can't create .../build/dequant_q4nx.o</code>) that broke fresh <br>clones.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2439/files#diff-df9125ffc6ce437fe924296c6404d5b9fa921303868bd2b694d67a6c59824f06">+8/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
## Correction to this PR's commit message

The commit message says `engine/npu/build` is **"not gitignored"**. That is wrong — it **is** gitignored,
by the general `build/` rule at `.gitignore:16`.

My check ran while the directory did not exist, and `build/` is a *directory-only* pattern, so git had
nothing to match and reported nothing:

```
# while the dir did not exist:
$ git check-ignore -v engine/npu/build     -> (no output)
# now that it does:
$ git check-ignore -v engine/npu/build     -> .gitignore:16:build/    engine/npu/build
```

That silence meant *"the directory is not there"*, not *"the rule does not apply"* — the same shape as
the bug being fixed, where a missing directory produced a confusing failure instead of a clear one.

**The fix and the conclusion are unchanged**: a fresh clone lacks the directory either way, which is
why the `mkdir -p` had to move. Only the stated mechanism was wrong.

I tried to amend the commit rather than leave a false statement in the message, and the push was refused
by a protected-branch hook (`protected branch hook declined`) — correctly, since this repo does not
rewrite pushed history. So the correction lives here, and in the PR description, instead.

Worth noting for anyone relying on commit messages: the merge queue's squash **does** carry these
messages into `main`'s history (it concatenates the branch's commit messages under the PR title — see
`51cd226e5` and `380f0b90f`), so a wrong aside in one is not discarded.

Issue #2329, opened 2026-09-13
bug(npu): Qwen3-4B cannot initialize on main — the split-G artifact for IM=9728 is not in the repo (FAIL G)


bong-water-water-bong:
## Summary

On `main` @ `0a130f4d3`, `npu_engine_qwen3_4b` **cannot initialize**: the shape
selector puts Qwen3-4B on the split-GU path, and the artifact that path requires
is not in the repo — or anywhere on this box.

```
$ NPU_XCLBIN_DIR=$PWD/engine/npu/xclbins ./engine/npu/build/npu_engine_qwen3_4b \
    ~/.config/flm/models/Qwen3-4B-NPU2/model.q4nx 4 ids256.txt
H=2560 NC=36 NH=32 NKV=8 HD=128 IM=9728 NV=151936 GU_split=1 rope_theta=1000000
  I8Ctx: xclbin/kernel init failed: No such file
    '.../engine/npu/xclbins/final_i8_G_K2560_N9728.xclbin'
FAIL G
[exit 1]
```

## Why it is shape-selected, not a bad path

`npu-infer/include/model_config.h:268` picks the split when the FFN is wide:

```c
cfg.gu_split = (cfg.IM * 2 > 14336);      // IM = 9728 for Qwen3-4B -> split
```

The split path then wants a **G** context (not GU), sized from the model's own
dimensions (`npu_engine_universal.cpp:924`, `:970`), i.e. `final_i8_G_K2560_N9728.xclbin`.

`main` tracks these split-G artifacts:

```
engine/npu/xclbins/final_i8_G_K2048_N512.xclbin
engine/npu/xclbins/final_i8_G_K2560_N8192.xclbin
engine/npu/xclbins/final_i8_G_K2560_N9216.xclbin
engine/npu/xclbins/final_i8_G_K2560_N12288.xclbin
engine/npu/xclbins/final_i8_G_K3072_N8192.xclbin
```

There is no `N9728`. Qwen3-4B is the shape that needs it.

## Scope

The G context is initialised **before** any path selection, so this is not
route-specific — all three of these fail identically with `FAIL G`:

| route | flags | result |
|---|---|---|
| dense int8 per-op | (default) | `FAIL G` |
| whole-layer runlist | `NPU_RUNLIST=1` | `FAIL G` |
[...]

PR #2457, opened 2026-09-17
npu: give ip() the dimension-keyed fallback xp() already has (Qwen3-4B and Phi-4-mini instruction lookup)

### LINKED ISSUE #2456: bug(npu): ip() lacked xp()'s dimension-keyed fallback — 6 instruction lookups across Qwen3-4B and Phi-4-mini silently used the single-core-row generator
## Summary

`npu_engine_universal.cpp` resolves **two** artifacts per tensor slot: a compiled
xclbin (`xp()`) and an instruction file (`ip()`). `xp()` had a dimension-keyed
fallback; `ip()` did not. Any slot whose instruction file is committed under its
shape name rather than a model tag therefore found *no* instruction file and
silently dropped to the runtime instruction generator — which, per
`npu_engine_i8ctx_inc.h:71-103`, emits a **single-core-row** sequence whose own
comment says it will *silently compute the wrong result* against a multi-row (v27)
xclbin.

The two lookups also disagree about which name wins, so a slot can pair an xclbin
found by dimensions with instructions found by tag — or with generated ones.

## Observed symptom

Prefilling Qwen3-4B (split-G path, on `main` @ `ab956f860`):

```
No insts for QKV, using runtime generator
WARN: generating single-core-row instructions ...
generated N instr bytes
No insts for O, using runtime generator
WARN: generating single-core-row instructions ...
generated N instr bytes
```

Exactly two fallbacks — QKV and O — while G/D/U loaded their tag-named `.txt`
files normally. Per-layer prefill was ~63 s versus ~0.86 s for Qwen3-8B: **73x
slower while being 2x smaller**, which is what sent me looking.

## Root cause

```c
// xp() — engine/npu/src/npu_engine_universal.cpp:829
auto xp=[&](const char*t, int K, int N) -> std::string {
    ...tag loop...
    return xd+"/final_i8_"+t+"_K"+K+"_N"+N+".xclbin";   // dimension-keyed fallback
};

// ip() — before the fix
auto ip=[&](const char*t){
    ...tag loop...
    return base+"_"+cfg.model_tag+".txt";               // no dimension-keyed fallback
};
```

`init_i8()` uses the returned path only as a probe:

```c
std::string xp_s=xp(t,K,N), ip_s=ip(t);
FILE* f=fopen(ip_s.c_str(),"rb");
if(f){fclose(f); return ctx.init(dev,xp_s.c_str(),ip_s.c_str(),4,NC);}
fprintf(stderr,"  No insts for %s, using runtime generator\n",t);
return ctx.init_with_generator(dev,xp_s.c_str(),XM,K,N,NC);
```

so a miss is silent by construction: the xclbin still resolves by dimensions, and
only the instructions degrade.

## Blast radius (measured, not estimated)

Replaying the lookup against the real committed files in `engine/npu/xclbins/`,
using the engine's own tag derivation (`:562-584`, including the `-`/`.`/`\` -> `_`
normalisation and the `_npu2`/`_instruct`/`_it` suffix stripping) for the 18 model
directories actually deployed under `~/.config/flm/models/`:

| deployed model | runtime tag | slots that found no instructions |
|---|---|---|
| `Phi4-mini-Instruct-NPU2` | `phi4_mini` | QKV, O, G, D |
| `Qwen3-4B-NPU2` | `qwen3_4b` | QKV, O |

**6 instruction lookups across 2 shipped models.** The committed files were there
the whole time — e.g. Qwen3-4B's QKV shape (K=H=2560, N=(NH+2*NKV)*HD=6144) has
`insts_i8_QKV_K2560_N6144.txt`, and Phi-4-mini's have `insts_i8_*_K3072_N8192.txt`
etc. `ip()` simply never tried the dimension-keyed name.

Two different naming accidents feed the same hole: Qwen3-4B has no
`insts_i8_QKV_qwen3_4b.txt` at all, while Phi-4-mini's files are committed as
`..._phi4_mini_4b.txt` but its deployed directory derives the tag `phi4_mini`.

## Why it is worse than slow

`engine/npu/src/npu_engine_i8ctx_inc.h:71-103` warns that the generated sequence is
single-core-row and that "pairing it with [a v27 multi-row xclbin] silently
computes the wrong result". Whether Qwen3-4B/Phi-4-mini were actually computing
wrong values, or only running slowly, is **not yet measured** — the confirmed
effects so far are the startup warnings and the 73x prefill regression. Related
history: #1510 (closed) covered the v26 single-core vs v27 multi-row artifact
split.

## Fix

Give `ip()` the same dimension-keyed fallback `xp()` has, and pass the shape at
every call site (4 of them). PR to follow — it also adds
`Testing/npu_insts_lookup_selfcheck.sh`, which pins the parity, replays the lookup
against the real files, and carries a pre-fix negative control (it fails on
`origin/main`).

Scope note: this is the instruction-file half. #2329's separate `FAIL G` complaint
(the missing `final_i8_G_K2560_N9728.xclbin`) is already resolved on `main`.

bong-water-water-bong:
## What

`xp()` and `ip()` resolve the two artifacts of a tensor slot — a compiled xclbin
and an instruction file. `xp()` fell back to a dimension-keyed name; `ip()` did
not. A slot whose instructions are committed under their **shape** rather than a
model tag therefore found nothing and silently dropped to the runtime instruction
generator.

`init_i8()` uses the returned path only as a probe:

```c
std::string xp_s=xp(t,K,N), ip_s=ip(t);
FILE* f=fopen(ip_s.c_str(),"rb");
if(f){fclose(f); return ctx.init(dev,xp_s.c_str(),ip_s.c_str(),4,NC);}
fprintf(stderr,"  No insts for %s, using runtime generator\n",t);
return ctx.init_with_generator(dev,xp_s.c_str(),XM,K,N,NC);
```

The xclbin still resolves by dimensions, so **only the instructions degrade, and
they degrade silently.** The generator emits a single-core-row sequence whose own
comment (`npu_engine_i8ctx_inc.h:71-103`) says it will *silently compute the wrong
result* when paired with a multi-row (v27) xclbin.

## How it showed up

Prefilling Qwen3-4B on `main` @ `ab956f860` printed exactly two generator
fallbacks — QKV and O — while G/D/U loaded their tag-named `.txt` files normally.
Per-layer prefill was ~63 s against Qwen3-8B's ~0.86 s: **73x slower while being
2x smaller.** That discrepancy is what led here.

## Blast radius (measured, not estimated)

Replaying the lookup against the real committed files, using the engine's own tag
derivation (`:562-584`, including `-`/`.`/`\` → `_` normalisation and the
`_npu2`/`_instruct`/`_it` suffix stripping) for the 18 model directories actually
deployed under `~/.config/flm/models/`:

| deployed model | runtime tag | slots that found no instructions |
|---|---|---|
| `Phi4-mini-Instruct-NPU2` | `phi4_mini` | QKV, O, G, D |
| `Qwen3-4B-NPU2` | `qwen3_4b` | QKV, O |

**6 instruction lookups across 2 shipped models.** The files were committed all
along — Qwen3-4B's QKV shape (K=H=2560, N=(NH+2·NKV)·HD=6144) has
`insts_i8_QKV_K2560_N6144.txt`; `ip()` never tried that name. Two different
naming accidents feed one hole: Qwen3-4B has no `insts_i8_QKV_qwen3_4b.txt` at
all, while Phi-4-mini's files are committed as `..._phi4_mini_4b.txt` but its
deployed directory derives the tag `phi4_mini`.

## Change

`ip()` now mirrors `xp()` — tag loop, then the dimension-keyed name — and all
four call sites pass the shape (`:1023`, `:1041`, `:1103`, `:1859`). Tag-named
files still win, so existing identifiers keep their meaning; the fallback only
adds a name that was previously unreachable.

## Test

`Testing/npu_insts_lookup_selfcheck.sh` (wired into `Testing/run_all.sh`) pins the
parity. It needs no device, no XRT and no compiler.

It **fails on `origin/main`** — that is the point, and I checked it rather than
assuming:

```
$ bash Testing/npu_insts_lookup_selfcheck.sh    # against pre-fix source
  FAIL - ip() has no dimension-keyed fallback (regressed?)
  FAIL - an ip() call site passes no shape:
        1010: ... ip_s=ip(t);
        1028: ... gi=ip("G");
        1090:            std::string gi = ip("GUSILU_i4");
        1845: ... ip(t).c_str() ...
  ok   - pre-fix lookup misses exactly QKV,O (negative control: QKV,O)
2 check(s) failed
```

and passes after. Two layers, with different jobs: the source contract reads the
shipped `.cpp` and so catches a revert, while the replay re-implements the lookup
to validate the data (which committed names exist for which real dims) and to
prove the bug class was reachable — it cannot notice the C++ being reverted, and
the script says so. The arity detector was validated against the pre-fix revision
before I trusted it; its earlier form missed all four `ip(t)`-style call sites.

## Not yet verified

I have **not** measured whether Qwen3-4B / Phi-4-mini were computing *wrong*
values or merely running slowly — the confirmed effects so far are the startup
warnings and the 73x prefill regression. The device was held by other work
throughout; a before/after numerics + timing run is the next step and I will
report it here rather than leave it implied.

## Scope

Closes #2456. **#2329 stays open** — its separate `FAIL G` complaint (the missing
`final_i8_G_K2560_N9728.xclbin`) is already resolved on `main` by a different
change, and its remaining "make the failure say which artifact is missing"
suggestion is untouched here.

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- `ip()` now falls back to shape-named instruction files

- All four `ip()` call sites pass `K` and `N`

- Generator fallback warning now names the missing path

- New selfcheck script pins lookup parity in `run_all.sh`

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Slot["Tensor slot<br/>xclbin + instructions"] -- "xp(t,K,N)" --> Xp["Tag loop, then<br/>final_i8_<t>_K<K>_N<N>.xclbin"]
  Slot -- "ip(t,K,N) (fixed)" --> Ip["Tag loop, then<br/>insts_i8_<t>_K<K>_N<N>.txt"]
  Ip -- "miss (was silent)" --> Gen["Runtime generator<br/>single-core-row, wrong vs v27"]
  Ip -- "hit" --> Init["I8Ctx::init with<br/>matching artifact pair"]
  Check["Testing/npu_insts_lookup_selfcheck.sh"] -- "source + replay" --> Ip
  Check -- "invoked by" --> RunAll["Testing/run_all.sh"]
```

<tr>
  <td>
    <details>
      <summary><strong>npu_engine_universal.cpp</strong><dd><code>ip() gains xp()'s dimension-keyed fallback and shape arguments</code></dd></summary>
<hr>

<ul><li><code>ip()</code> signature becomes <code>ip(const char* t, int K, int N) -> std::string</code>, <br>mirroring <code>xp()</code>.<br> <li> After the progressive tag-stripping loop, it now returns <br><code>base+"_K"+K+"_N"+N+".txt"</code> instead of <code>base+"_"+cfg.model_tag+".txt"</code>, so <br>shape-committed instruction files are reachable.<br> <li> All four call sites updated to pass the shape: <code>init_i8()</code> (<code>:1023</code>), <br><code>diag_split_g()</code> (<code>:1041</code>), the GUSILU_i4 pair (<code>:1103</code>), and the MoE <br><code>moe_ctx</code> lambda (<code>:1859</code>).<br> <li> The runtime-generator fallback warning is expanded to name the missing <br>file and state that single-core-row instructions silently compute the <br>wrong result against a multi-row (v27) xclbin, referencing <code>#2456</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2457/files#diff-63677a636d5ff19712cf034cc1e2c15a96b194137caee54d8bd6883b0b76a10b">+27/-7</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>npu_insts_lookup_selfcheck.sh</strong><dd><code>Self-check pins NPU instruction lookup parity and fallback</code></dd></summary>
<hr>

Testing/npu_insts_lookup_selfcheck.sh

<ul><li>New device-less self-check that verifies the <code>xp()</code>/<code>ip()</code> artifact-lookup <br>parity in <code>engine/npu/src/npu_engine_universal.cpp</code>.<br> <li> Layer 1 asserts the source contract: <code>ip()</code> and <code>xp()</code> both contain the <br><code>_K"+"_N"</code> dimension-keyed fallback, and no <code>ip()</code> call site omits its <br>shape (a one-arg <code>ip(</code> is flagged).<br> <li> Layer 2 replays the lookup in Python against the real <br><code>engine/npu/xclbins/</code> tree, parsing model dims from <code>npu_dims.h</code> rather <br>than hardcoding them.<br> <li> Carries a pre-fix negative control (pre-fix lookup must miss exactly <br><code>QKV,O</code> for <code>qwen3_4b</code>), asserts the fix resolves them, that no other <br>slot's resolution moves, and that tag-named files still win for <br><code>qwen3_8b</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2457/files#diff-1a611a3fdeea71f556d8d431b8f4f52aa60c4e3025970198f09c8cc13a273aec">+227/-0</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>run_all.sh</strong><dd><code>run_all.sh runs the new NPU lookup self-check</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/run_all.sh

<ul><li>Adds an "NPU insts/xclbin lookup parity" step to the aggregate test <br>runner.<br> <li> Runs <code>Testing/npu_insts_lookup_selfcheck.sh</code> with <code>PYTHON</code> forwarded, <br>counts it in <code>total</code>, and increments <code>fail</code> while printing the last 12 <br>lines on failure.<br> <li> Notes the check needs no device and no compiler and has a built-in <br>pre-fix control.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2457/files#diff-60779953d5d72f0d8ef3aea52f86b0b365bbdb42457aae2e5955d6320e32eb2c">+17/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
## Corrected census, and one correction to my own numbers

I extended the replay to every deployed model and found two things worth
recording.

**1. The 6-slot claim holds, independently.** Recomputing from scratch, honouring
`gu_split = (IM*2 > 14336)` so a fused model is asked for `GU` rather than a
`G`/`U` pair it never had:

```
fixed by this PR: 6 slots
  Phi4-mini-Instruct-NPU2/QKV, /O, /G, /D
  Qwen3-4B-NPU2/QKV, /O
```

**2. 20 more slots still fall back — by design, and I am not claiming them.**
Gemma3-1B (all 4), Gemma3-4B (all 5), Llama-3.1-8B (all 5), Qwen3-1.7B (all 5)
and Phi4-mini's `U` have **no** instruction file committed for their shape, under
any name. The generator is the documented, intended fallback for exactly that
case ("makes any model with compatible GEMM shapes work without pre-compiling
per-model instruction files"), so those are not lookup failures and this PR does
not change them. They are now at least loud rather than silent, which is the
second commit.

One caveat on the whole census: it derives the runtime tag the way the engine
does from the **model path** (`:562-584`). A caller that passes `--model-tag`
explicitly would derive a different tag. Qwen3-4B's case is confirmed
empirically — the pre-fix startup log shows exactly the two fallbacks (QKV, O)
that the static replay predicts — but the other rows are static predictions, not
observed runs.

## A number I got wrong, discarded

I also tried a census of the **xclbin** side and it reported 25 slots with no
xclbin. That verdict is wrong and I am not using it: `xp()` is not the only
xclbin source (there is the FLM per-model directory fallback around `:913`), and
the first version of that census ignored `gu_split` entirely, so it demanded a
`final_i8_G_*` for models that correctly use `GU`. Anyone reading that table
would have concluded that Gemma3 and Llama-3.1 cannot load, which is false. The
instruction-file side has no such second source — `ip()` only ever looks in
`engine/npu/xclbins/` — so that census is the one I trust.

PR #2467, opened 2026-09-17
ci(smoke): the HIP-kernel path filter matched zero files

bong-water-water-bong:
## The bug

The End-to-End Smoke Test gates on path filters, and one arm was inert:

```yaml
- "src/backend_*.hip"     # matches 0 files
```

There is no file by that name. Checked against the tree:

| pattern | matches |
|---|---:|
| `src/backend_*.hip` (before) | **0** |
| `src/*.hip` | 34 |
| `kernels/*.hip` | 47 |

The kernels are not named `backend_*.hip`. They live in `src/` and `kernels/`, and **both are compiled into the product** — `CMakeLists.txt:403-430` lists `src/ternary_gemv_launchers.hip`, `src/kv_cache_attn*.hip`, `src/sherry_gemv.hip`, `src/mamba2_kernels.hip`, `kernels/ternary_gemv_*.hip`, `kernels/bonsai_q1_gemv_soa.hip` and more.

So a **kernel-only** change — a HIP bugfix touching `src/hip_1bp_kernels.hip` or any `kernels/*.hip` — did not trigger this gate. That matters more than usual here because this workflow runs on `[self-hosted, strix-halo]`, i.e. it is the end-to-end test on real hardware.

## The change

```diff
-      - "src/backend_*.hip"
+      - "src/**.hip"
+      - "kernels/**.hip"
```

81 compiled kernel sources that the gate previously ignored.

## Verification

- The workflow still parses as YAML, with **9** path entries.
- **Every** remaining arm matches at least one file — the old arm was the only zero, so no other filter is inert:

| arm | matches |
|---|---:|
| `tools/unified_server.cpp` | 1 |
| `src/backend_*.cpp` | 27 |
| `src/**.hip` | 34 |
| `kernels/**.hip` | 47 |
| `src/zaya_engine.cpp` | 1 |
| `src/zamba2_engine*` | 3 |
| `src/gguf_*.cpp` | 4 |
| `CMakeLists.txt` | 1 |
| `.github/workflows/*` | 32 |

I checked the other workflows for the same class of defect and found none: `ci.yml` has no path filter and runs on every push/PR, and only this file had a `.hip` arm.

## One honest limitation

This PR also modifies `.github/workflows/*`, which is itself a path arm — so the smoke test will run for this change **regardless** of the HIP arms. The new `.hip` coverage is therefore evidenced by matching the patterns against the tree, not by watching a kernel-only PR trigger it. A genuine end-to-end demonstration would need a `.hip`-only change.

CI configuration only; no product code touched.

___

### **PR Type**
Bug fix

___

### **Description**
- Fix inert HIP path filter in smoke test

- Replace `src/backend_*.hip` with two real arms

- Now covers `src/**.hip` and `kernels/**.hip` kernels

- Kernel-only changes now trigger self-hosted gate

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["path filter: src/backend_*.hip"] -- "matched 0 files" --> B["inert gate"]
  B -- "replaced with" --> C["src/**.hip (34 files)"]
  B -- "replaced with" --> D["kernels/**.hip (47 files)"]
  C --> E["End-to-End Smoke Test runs on kernel-only changes"]
  D --> E
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Configuration changes</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>end-to-end-smoke.yml</strong><dd><code>Replace dead HIP path filter with real kernel globs</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/end-to-end-smoke.yml

<ul><li>Removed the inert <code>src/backend_*.hip</code> path filter (matched 0 files).<br> <li> Added <code>src/**.hip</code> and <code>kernels/**.hip</code> arms covering 81 compiled kernel <br>sources.<br> <li> Added an explanatory comment citing <code>CMakeLists.txt:403-430</code> for kernel <br>locations.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2467/files#diff-ce7583116dfeff04a5365187d78ae67293490e66e36673d3233b26e96f97023c">+6/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
## Supporting evidence: the gate this fixes is demonstrably live

The PR above argues the inert arm mattered because this workflow runs on `[self-hosted, strix-halo]`. Here is the measured liveness, so the claim does not rest on the `runs-on` line alone:

```
End-to-End Smoke Test — last 20 runs:
  success      19
  in_progress   1   <- this PR
```

Recent completed runs, as examples:

```
2026-09-17T04:17  success   fix/2420-spike-decision-uses-clones-pe
2026-09-17T03:56  success   fix/2420-spike-decision-uses-clones-pe
2026-09-15T11:15  success   fix/2411-search-index-claims
2026-09-15T10:39  success   fix/2408-docs-census-claims
```

A runner that no longer existed would leave these queued, not green. So the missing `.hip` coverage was a **real hole in a gate that executes regularly on real hardware** — not a formality on a dormant workflow.

### The same audit over the other 31 workflows found nothing else

I checked every workflow for this class of defect rather than only fixing the one I noticed:

- **Inert path filters:** `src/backend_*.hip` was the only pattern in the repo matching zero files. Every other arm matches at least one.
- **Missing entrypoints:** three apparent hits, none real. Two are wrapped in `if [ -f … ]` with explicit *"not present — skipping"* notices (`scripts/generate_benchmark_table.py`, `tools/gen_numbers.py`, both deliberately retired), and one was my own grep matching the **destination of a `cp`** — `release.yml:180` creates `dist/run.sh` from `packaging/tarball-run.sh`, so it is not an invocation at all.
- `ci.yml` has no path filter and runs on every push/PR, so `.hip` changes were never wholly uncovered — they simply never reached *this* end-to-end gate.

### Unchanged caveat

The smoke test runs for this PR because it modifies `.github/workflows/*`, which is itself an arm — not because of the HIP arms. The `.hip` coverage is evidenced by matching patterns against the tree (34 + 47 files) and by the gate's liveness above, not by watching a kernel-only PR trigger it. A true end-to-end demonstration still needs a `.hip`-only change.

Issue #2451, opened 2026-09-17
bug(npu): the STD attention q/gate pack on main drops the *H row stride — wrong q and gate for every non-zero head


bong-water-water-bong:
## The defect, on `main`

`engine/npu/src/npu_engine_universal.cpp` packs the STD fused attention q+gate weights with a **row offset
that is missing its `* H`**:

```
origin/main:1419
    transpose_pack(qkv_w + gdn_k_off, gdn_k_off, H, w.data(), t, gdn_k_off);            // K
```

`qkv_w` is row-major `[rows][H]`, and `transpose_pack` reads `src[o*in_f + i]` with `in_f == H`. So the row
offset has to be scaled by `H`, and it is not — the pointer advances by `gdn_k_off` **floats** instead of
`gdn_k_off * H`. Head 0 is right by accident (offset 0); every other row reads a sliding window inside the
first couple of rows of `q_proj`.

The GDN branch 20 lines above does scale by `H`, which is why the error profile steps up at each
full-attention layer while the GDN layers look fine — the asymmetry that made this look like an
attention/O-projection bug for several rounds.

## What it costs

Measured by the lane that found it (Round 99, handback + commit `8c4c47a90` in `~/tip-wt`), L3 position 0,
against the engine's own dumped input:

| | before | after |
|---|---|---|
| `corr(engine_q, ref_q)` | **+0.0820** (≈1/16 = head 0 only) | **+0.999876** |
| `corr(engine_gate, ref_gate)` | **+0.0275** (≈0) | **+0.999871** |
| per-layer rel err L3 / L39 | 0.5403 / 0.9274 | 0.1315 / 0.2023 |

## The fix exists and is **not on any remote**

I verified this rather than assuming it. From the shared clone, after
`git fetch origin '+refs/heads/*:refs/remotes/origin/*'` (136 remote branches):

```
git branch -r --contains 8c4c47a90   -> nothing
8b59ead9c:1975  transpose_pack(qkv_w + (size_t)gdn_k_off * H, gdn_k_off, H, w.data(), t, gdn_k_off);
```

So the corrected form is `qkv_w + (size_t)gdn_k_off * H`, it is committed, and it is **unreachable from
origin** — it lives in a detached worktree. I have flagged that separately; this issue exists so the defect is
tracked whether or not those commits are pushed, because right now `main` ships the bug and nothing records it.
[...]

PR #2435, opened 2026-09-16
NPU: dual-engine concurrency beats FLM; four Qwen3-constant bugs fixed

bong-water-water-bong:
### **User description**
Draft, because this branch is 1286 commits ahead of `main` and has never had a PR. Opening
it so the work is visible and reviewable; it is **not** a request to merge today.

## Why now

Until this week the branch was a **17,347-file, 2.3 GB** change — 16,998 of those files were
generated capture artifacts contributing 88 insertions between them. That is why no PR was
ever opened. `npu-infer/captures/` is now untracked (`a1f96ff13`, `8e4580a91`) and the branch
is a **367-file** review.

## Headline: the dual NPU engine beats FLM on aggregate throughput

The NPU runs multiple engines concurrently at full per-stream speed; FLM's server queues.

| configuration | aggregate | per-stream loss |
|---|---|---|
| 2 x Qwen3-0.6B | **156-158 tok/s (2.05-2.08x)** | none |
| 3 x Qwen3-0.6B | **227 tok/s (2.99x)** | none (76 -> 74) |
| 0.6B + 4B (mixed) | 98 tok/s (~1.01x of the two solos) | none |
| **FLM, 2 concurrent** | **7.1 tok/s (1.00x)** | **it serializes** |

Every single-stream comparison in the repo understated the engine, because FLM only serves
one stream. Measured four ways; see `benchmarks/RESULTS-default-path-parity-verification-2026-09-16.md`.

## Bug fixes, each with a reproduced symptom and a verified fix

Four of these were **one class: a Qwen3 constant applied to every model.** All four produced
the *same* timeout/divergence symptom, which is why they read as one mystery rather than four
bugs, and each fix exposed the next because every site sits earlier than the previous one's
success.

| commit | bug |
|---|---|
| `91cfc0fd6` | `sess_model_dir(H)` picked a Qwen3 `layer.xclbin` for non-Qwen3 models — Nanbeige (H=2560) got Qwen3-4B's, Phi-4 (H=3072) fell through to 0.6B's. First runlist forward then timed out. Nanbeige 0.14 -> **27 tok/s** (FLM 21), Phi-4 -> **24 tok/s** (FLM 19). |
| `4a269430e` | the ELF cache was seeded from another model's shipped set — Nanbeige's cache got 4,249 symlinks to Qwen3-4B's ELFs. |
[...]

PR #2479, opened 2026-09-17
fix(npu): the fused QKV pack dropped the *H row stride — Q was re-packed as K and V

bong-water-water-bong:
## The defect

Closes the code-level half of #2451. `engine/npu/src/npu_engine_universal.cpp`, `struct Bf16Ctx`, the two branches that pack a **fused** qkv tensor. `qkv_w` comes from `dq()`, which returns row-major `[out_features, in_features]`, and `transpose_pack` reads `src[(size_t)o * in_f + i]` with `in_f == H`. Selecting the output-row block starting at row `R` therefore needs `src + R * H`. Four sites advanced by `R` **floats**:

| line | branch | was | is |
|---|---|---|---|
| 1439 | GDN K | `qkv_w + gdn_k_off` | `qkv_w + (size_t)gdn_k_off * H` |
| 1440 | GDN V | `qkv_w + gdn_v_off` | `qkv_w + (size_t)gdn_v_off * H` |
| 1489 | STD q | `qkv_w + h*2*hd` | `qkv_w + (size_t)(h*2*hd) * H` |
| 1490 | STD gate | `qkv_w + h*2*hd + hd` | `qkv_w + (size_t)(h*2*hd + hd) * H` |

## Independent verification

Not eyeballed — simulated. `transpose_pack` copied verbatim, source filled with `value == row index`, Qwen3.6 geometry (`H=2048`, `k_off=2048`, `v_off=4096`; `nh=16`, `hd=128`), so the min/max value in a destination block names exactly which source rows it read:

```
GDN K      OLD reads rows    1..2048    NEW reads 2048..4095  ✅
GDN V      OLD reads rows    2..4097    NEW reads 4096..8191  ✅
STD q h=0  OLD reads rows    0..127     (right by accident — offset 0)
STD q h=1  OLD reads rows    0..128     NEW reads  256..383   ✅
STD q h=15 OLD reads rows    1..129     NEW reads 3840..3967  ✅
STD gate   OLD reads rows    0..128     NEW reads the head's gate rows ✅
```

The GDN branch was packing **Q as K and Q/K as V**. In the STD branch only head 0's q was right, which is precisely why the original symptom looked like `corr(engine_q, ref_q) ≈ 1/16`.

This also **corrects the existing fix commit's own claim**. `8c4c47a90` says *"all 14 call sites were enumerated by text search and only these two lacked the offset scaling"* — but on `origin/main` there is no `*H`-scaled site at all (`git grep -nE 'transpose_pack\([^;]*\* *H' origin/main` → 0 hits), so GDN was defective here too. That commit's message was true of *its* base, which already carried the GDN half.

## Scope

`dst_offset` is deliberately untouched, so the packed layout stays `[all q heads | all gate heads]` and **no converter change is needed**.

Siblings checked and unaffected — `npu_engine_cb.cpp`, `spec-decode/engine/npu_target_model.h` and `final_integrate.py` all pass *separate* `qw`/`kw`/`vw` tensors at offset 0. `transpose_pack` has 15 call sites; only these four pass a non-zero source offset.

## Relationship to the in-flight work

#2451 records a fix in `~/tip-wt` (`8c4c47a90`) that is reachable from **no ref, local or remote**, and rides a branch 1337 commits ahead of a 16-behind base. Its GDN half is identical to this one. Its STD half is a **different shape** — a single `transpose_pack(qkv_w, t, H, w.data(), t, 0)`, which pairs with a converter that already emits `[q_all | gate_all]` (`:2340` carries that comment) — so it is not a two-hunk cherry-pick, and that is why this takes the offset form instead.

This is the minimal correction that makes `main`'s *existing* layout correct. If that branch lands, the GDN half is the same and the STD half supersedes this one.

## Not verified here

The `corr` figures in #2451 are the finding lane's measurement — I did not re-run them (the NPU path is device-gated). What this PR adds is the source-level proof and an executable simulation of the packing geometry. No end-to-end 35B claim.

___

### **PR Type**
Bug fix

___

### **Description**
- Fix missing `* H` row stride in fused QKV pack

- GDN branch K/V blocks read Q rows instead

- STD branch only head 0's q was correct

- Add comments documenting the row-major invariant

___

### Diagram Walkthrough

```mermaid
flowchart LR
  QKV["qkv_w: row-major [rows][H]"] -- "src + R floats (wrong)" --> OLD["transpose_pack reads rows 0..n window"]
  QKV -- "src + R * H floats (fixed)" --> NEW["transpose_pack reads block at row R"]
  OLD -- "GDN: Q packed as K, Q/K as V" --> BAD1["GDN attention wrong"]
  OLD -- "STD: only head 0 q correct" --> BAD2["STD q/gate wrong for h>0"]
  NEW --> OK["All heads packed correctly"]
```

<tr>
  <td>
    <details>
      <summary><strong>npu_engine_universal.cpp</strong><dd><code>Fix `* H` row-stride offset in both fused QKV pack branches</code></dd></summary>
<hr>

<ul><li>In <code>struct Bf16Ctx</code>, the GDN fused-QKV branch now advances the source <br>pointer by <code>(size_t)gdn_k_off * H</code> and <code>(size_t)gdn_v_off * H</code> instead of <br><code>gdn_k_off</code>/<code>gdn_v_off</code> floats, so K and V read their own output-row <br>blocks rather than a window inside <code>q_proj</code>'s first rows.<br> <li> The STD fused-q/gate branch scales both head pointers by <code>H</code>: <code>qkv_w + </code><br><code>(size_t)(h * 2 * std_hd[l]) * H</code> for q and <code>qkv_w + (size_t)(h * 2 * </code><br><code>std_hd[l] + std_hd[l]) * H</code> for gate, fixing every non-zero head.<br> <li> Destination offsets (<code>t</code>, <code>gdn_k_off</code>, <code>gdn_v_off</code>, <code>h * std_hd[l]</code>, <code>std_nh[l] </code><br><code>* std_hd[l] + h * std_hd[l]</code>) are deliberately unchanged, preserving <br>the packed <code>[all q heads | all gate heads]</code> layout the converter <br>expects.<br> <li> Adds explanatory comments in both branches documenting the row-major <br><code>[out_features, in_features]</code> invariant of <code>qkv_w</code> and referencing issue <br>#2451.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2479/files#diff-63677a636d5ff19712cf034cc1e2c15a96b194137caee54d8bd6883b0b76a10b">+16/-4</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
The GDN half and the simulation method are solid — `transpose_pack` copied verbatim with `value == row index` is the right way to show which rows a call selects. But the STD hunk rests on a premise I think the tree refutes, and it is worth settling before merge, because the failure mode here is that the *symptom* goes away while q/gate stay wrong.

## The premise

> "This is the minimal correction that makes `main`'s *existing* layout correct."
> "a single `transpose_pack(qkv_w, t, H, w.data(), t, 0)`, which pairs with a converter that already emits `[q_all | gate_all]`"

That reads the restructure as the fix for a *different* converter, and main's loop as correct for the layout it actually receives. The only producer of these bundles in this tree says the opposite:

`third_party/FLM_Q4NX_Converter/q4nx/models/qwen35.py:53-61`

```python
if layer_id % full_attntion_interval == (full_attntion_interval - 1):
    if "q_proj" in self.forward_name_map[gguf_tensor.name]:
        DH = self.gguf_reader.fields["qwen35.attention.value_length"].contents()
        d, m, qw = unpacked
        d = rearrange(d, '(g p h) c -> (p g h) c', p = 2, h = DH).contiguous()
```

`(g p h) -> (p g h)` with `p = 2` fires on **full-attention layers** — the STD branch — and moves the q/gate axis to the front, i.e. rows become `[q_all (g·hd) | gate_all (g·hd)]`. The comment above the loop (`rows [h*2*hd, h*2*hd+hd) = q, [+hd, +2*hd) = gate`) describes what the loop *assumes*; it is the same assumption that produced the bug, not an independent record of the file.

## And it is not only the converter reading

`~/tip-wt` `8c4c47a90` — the commit whose measurement #2451 reports — carries the restructure, not an offset, and no loop at all:

```
$ git -C ~/tip-wt show 8c4c47a90:engine/npu/src/npu_engine_universal.cpp | grep -n 'transpose_pack(qkv_w' | tail -1
2078:        transpose_pack(qkv_w, t, H, w.data(), t, 0);   // file rows are already [q_all | gate_all]
```

#2451's table gives `corr(engine_q, ref_q) 0.0820 -> 0.999876` "measured by the lane that found it (commit `8c4c47a90`)". The GDN half cannot move `corr(q)` — that block packs K and V — so the jump is the STD change, and the STD change in that commit packs the file's rows **verbatim**. A verbatim pack reaching 0.9999 is a measurement of the file's row order: had the file been `[q_h, gate_h]`-interleaved, packing it in order would have left `corr(q)` where it was.

## What the offset form does under that layout

For head `h`, the source rows it reads are `h*2*hd` (as q) and `h*2*hd + hd` (as gate) — i.e. it takes `q_{2h}` and `q_{2h+1}`. Distinct rows, so the sliding window disappears; still not this head's q or gate. That is the part I would not want to land, because the numbers improve and the defect stops being visible.

## What is not in dispute, and what would settle it

* The GDN hunk is right, and it is byte-identical to #2515 — whichever of the two merges second should drop its GDN hunk. #2515 additionally carries `Testing/npu_pack_stride_selfcheck.py`, which **exempts the STD sites precisely because of this ambiguity** and fails when the exemption goes stale, so if this PR lands as-is that guard will go red by design and I will rebase it.
* Neither form touches the third thing #2451 records: the QKV context is sized from `cfg.qkv_total` (`npu_engine_universal.cpp:1999`) while this branch packs `t = std_nh[l]*std_hd[l]*2` rows, which is what the segfault note is about for Qwen3.5-4B. Worth folding in whichever way the row question goes.
* Settle it with the method already in this PR: run the same simulation with the converter's row order, or re-run the lane's `corr(q)/corr(gate)` check against the offset form. If the measurement turns out to have been taken with the offset form rather than `8c4c47a90`'s restructure, my reading is wrong and I will withdraw it — the pivot is which code produced `0.999876`.

bong-water-water-bong:
Queue prediction, from `git merge-tree` rather than from reading the diffs.

#2515 (`fix/npu-transpose-pack-row-stride`) is **ahead of this one in the queue** — pos 10 against this entry's pos 20 — and the two are **not duplicates**: they fix different bugs in the same function, and each carries 198 lines the other does not.

They conflict:

```
$ git merge-tree --write-tree origin/fix/npu-std-gdn-row-stride origin/fix/npu-transpose-pack-row-stride
CONFLICT (content): Merge conflict in engine/npu/src/npu_engine_universal.cpp
1
```

each against current main:

```
fix/npu-std-gdn-row-stride        rc=0  CLEAN
fix/npu-transpose-pack-row-stride rc=0  CLEAN
```

So both are individually landable and mergeable with main, and they only collide **with each other**. #2515 merges first, and when this entry's turn comes its speculative merge will hit that same conflict and the queue will eject it — which loses a complementary fix rather than a duplicate one.

The resolution is a rebase onto #2515's merge (not a close): both changes want to live in the same region of `npu_engine_universal.cpp`, so the combined file needs both. If it helps, I can prepare that rebase and push it as a separate branch the moment #2515 lands — say the word, or do it yourself and I will leave it alone.

bong-water-water-bong:
Closing as superseded by #2515, which landed the identical GDN half on `main`.

## Why this is a close and not a rebase

**The GDN half is already on `main`.** #2515's two `transpose_pack` hunks in `engine/npu/src/npu_engine_universal.cpp` are byte-identical to this PR's:

```c
transpose_pack(qkv_w + (size_t)gdn_k_off * H, gdn_k_off, H, ...);  // K
transpose_pack(qkv_w + (size_t)gdn_v_off * H, gdn_v_off, H, ...);  // V
```

So a rebase would leave only the STD half — which is the half #2515 deliberately excludes.

**The remaining STD half is now actively harmful, not just redundant.** #2515 added `Testing/npu_pack_stride_selfcheck.py`, which `Testing/run_all.sh` invokes (the required `Self-checks` job). It carries a staleness clause: every `KNOWN_UNSCALED` exemption must still match a live call site, or the check fails. Its exemption for the STD sites states the intended fix outright:

> *"STD q block, per-head interleave; the correct fix is the `[q_all | gate_all]` restructure from the converter, **not an offset scaling** (#2451)"*

Reproduced on `main` + this PR's STD change applied:

```
$ python3 Testing/npu_pack_stride_selfcheck.py
VIOLATION:
  - the KNOWN_UNSCALED exemption 'qkv_w + h * 2 * std_hd[l]' no longer matches a
    call site: the site was fixed or rewritten, so delete the exemption
    (STD q block, per-head interleave; the correct fix is the [q_all | gate_all]
     restructure from the converter, not an offset scaling (#2451))
  - the KNOWN_UNSCALED exemption 'qkv_w + h * 2 * std_hd[l] + std_hd[l]' no longer
    matches a call site: ...
exit 1
```

Baseline on unmodified `main` is `exit 0`. So landing this PR would turn a required CI job red on `main`.

## Independent corroboration

`goal/runlist-decode-wire` (PR #2435) independently reaches the same verdict — it carries the GDN `* H` scaling and leaves the STD sites unscaled, matching #2515 and contradicting the offset-scaling form here.

## Where the STD work goes

Not closed as "wrong work" — the STD sites do need fixing. Per #2515 and this PR's own analysis, the correct form is the `[q_all | gate_all]` restructure tracked by #2451, not `* H` on `h * 2 * std_hd[l]`. This PR's STD reasoning is preserved in the thread above for whoever picks up #2451.

For the record, the cross-file triage: #2479 and #2515 could never both land — identical GDN hunks on the same lines — so one had to be closed. #2515 carries the guard and the rationale, and is already merged.

Issue #2307, opened 2026-09-13
bug(npu): the non-fused p1/p2 split int4 path is unhealthy — silent rms=4.6e8 / corr=-nan (from #2282)


bong-water-water-bong:
## Summary

Filed from the closing notes of #2282 (merged as `f3825fbb6`), which is where this finding currently lives — in a merged PR body, where the next person to touch the int4 MoE path will not find it.

While A/B-ing the rebuilt int4 fused xclbin, the **non-fused (p1/p2 split) int4 path** was measured as broken, on both artifacts:

| artifact | non-fused p1/p2 split behaviour |
|---|---|
| committed `final_i8_MOE_GUSILU_i4_zaya.xclbin` (67,306 B, pre-#2282) | NPU writes nothing — `npu rms=0.0000`, `corr=-nan` |
| rebuilt (75,040 B, post-#2282) | emits `rms=4.6e8` |

The fused path (`NPU_FUSED=1`) is the healthy one and is what #2282 gated on (byte-identical `[MoE L1 fused dbg]` diagnostics between the two artifacts).

## What is and is not established

- **Established:** the numbers above, from the engine's own L1 diag line, same session/environment, D stage unchanged.
- **Not established:** whether the post-#2282 *committed* artifact still shows the first row — that row was measured against the old committed xclbin, and `main` now ships the rebuilt one. A fresh measurement on `f3825fbb6` is the first step.
- **Trap when reproducing** (from #2282): `NPU_FUSED_I4=1` alone silently takes the *non-fused* path and reports `corr=-nan`; the parent switch `NPU_FUSED=1` is required. Easy to "reproduce" the fused path and conclude the wrong thing.

## Suggested scope

1. Re-measure the split path on `f3825fbb6` and confirm which of the two rows is current.
2. If it is still broken, decide the intent: is p1/p2 split a supported fallback, or dead code that should fail loudly rather than emit `rms=4.6e8`? A silent wrong-output path is worse than a refused one.

No owner yet; flagged rather than fixed.

bong-water-water-bong:
## The split path is selected by model shape, not by a flag — so this is not dead code

The issue asks whether p1/p2 split is a supported fallback or dead code that should be deleted. It is **auto-selected from the artifact's own dimensions**:

`npu-infer/include/model_config.h:268`
```c
cfg.gu_split = (cfg.IM * 2 > 14336);
```
[...]

PR #2483, opened 2026-09-17
tool(npu): a reproducible xclbin census for #2329 — 33 slots have no artifact

bong-water-water-bong:
### **User description**
Part of #2329 — **the xclbin half of the census**, which that issue left open.

The instruction-file half was done. The xclbin half was withdrawn by its author as untrustworthy: the first attempt ignored the two things that make `xp()` non-obvious, so it demanded `final_i8_G_*` from models that correctly use `GU` and concluded Gemma3 and Llama-3.1 "cannot load at all". A census that says that is worse than no census, so this one takes both from the source.

## What it models

| | source |
|---|---|
| **shapes** | `npu-infer/include/model_config.h:262-279` — `QKV K=H N=NH*HD+2*NKV*HD` · `O K=NH*HD N=H` · `G,U K=H N=IM` when `gu_split = (IM*2 > 14336)` · `GU K=H N=IM*2` otherwise · `D K=IM N=H` |
| **lookup** | `npu_engine_universal.cpp:829-844` — three candidates in order; the last is returned **without checking it exists** |
| **tag** | `…:562-584` — basename minus extension, parent dir if that is `model`, lowercased, `-`/`.`→`_`, one suffix off `{_npu2,_instruct,_it,_it_npu2}` |

That third candidate is why this **reports** rather than asserts: `xp()` hands back a path it never checked, and the failure surfaces later as the context init failing. A slot is only "present" here if a candidate file actually exists.

## Result — 85 slots checked, **33 with no xclbin**

**All slots resolve (11):** Gemma4-E2B, Gemma4-E4B, Llama-3.2-3B, Nanbeige4.1-3B, Phi4-mini, Qwen3-0.6B, Qwen3-4B, Qwen3-8B, Qwen3-VL-4B, Qwen3.5-4B, Qwen3.6-35B-A3B

**Missing (7):** Gemma3-1B (4), Gemma3-4B (5), LFM2-1.2B (5), LFM2-2.6B (5), **Llama-3.1-8B (5)**, Llama-3.2-1B (5), Qwen3-1.7B (4)

## Why the numbers should be believed

1. **The shape formulas reproduce the one case the engine records.** Qwen3-4B (H=2560 NH=32 NKV=8 HD=128 IM=9728) → `final_i8_G_K2560_N9728.xclbin`, the exact artifact #2329 was filed about.
2. **The tag rule reproduces `qwen3_4b`**, and the census then finds that model's G/U/D **via tag** and its QKV/O **via dim** — the mixed resolution the `ip()` comment warns about, where two halves of one context resolve from different names.
3. **Every MISSING verdict was spot-checked by hand.** `final_i8_QKV_K1152_N1536`, `final_i8_D_K6912_N1152`, `final_i8_GU_K1152_N13824` are genuinely absent; Qwen3-0.6B's four are genuinely present. Gemma3, Qwen3-1.7B and LFM2 have **no artifact of any name**, which independently reproduces the instruction-half result for Llama-3.1-8B.

**Llama-3.1-8B is the instructive one:** its tag is `llama_3_1_8b`, and `xp()` strips **leading** underscore groups only (`llama_3_1_8b → 3_1_8b → 1_8b → 8b`), so it can never reach the committed `final_i8_G_llama.xclbin`. That is exactly the gap #2463 closes by renaming the artifacts to carry the full identity.

## Deliberately not wired into `run_all.sh`

The missing artifacts are a known existing condition. A census that fails CI would be the same over-reach that got the first one withdrawn — it would demand someone fix seven shipped models' artifacts before any unrelated PR could land.

## Scope (so the table isn't read as more than it is)

The **legacy per-op path** (`init_i8`). It does not model `--use-flm-xclbin` (`mm.xclbin`), the i4/`GUSILU_i4` path, or the instruction half (`ip()`), which #2329 censused separately. Shapes come from each model's `config.json` rather than the engine's tile-metadata derivation — validated against the Qwen3-4B case above — and a model whose config cannot supply the five dims is reported as **underivable** rather than guessed at (all 18 are derivable today).

___

### **PR Type**
Tests, Enhancement

___

### **Description**
- Adds `Testing/npu_xclbin_census.py`, a reproducible artifact census tool

- Models engine lookup order from source: tag, stripped tag, `K_N`

- Reports slots missing xclbins and slots missing insts files

- `--pairing` mode audits committed xclbins against `insts_i8_*.txt`

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Testing/npu_xclbin_census.py<br/>new census tool"] -- "shape formulas" --> B["model_config.h<br/>gu_split = IM*2 > 14336"]
  A -- "replicates xp()/ip()" --> C["lookup order<br/>tag, stripped tag, K_N"]
  C -- "no file found" --> D["slot MISSING xclbin<br/>engine prints FAIL slot"]
  C -- "xclbin but no .txt" --> E["generator fallback<br/>N insts all with a single core.row"]
  A -- "--pairing" --> F["committed artifact audit<br/>RISK vs note classification"]
  A -- "--json" --> G["machine-readable report"]
```

<tr>
  <td>
    <details>
      <summary><strong>npu_xclbin_census.py</strong><dd><code>New xclbin census tool modeling engine lookup and shapes</code>&nbsp; </dd></summary>
<hr>

Testing/npu_xclbin_census.py

<ul><li>New 306-line reporting script that censuses which deployed models have <br>an xclbin and instruction file for every slot the legacy <code>init_i8</code> path <br>requests.<br> <li> Reproduces the engine's shape formulas from <code>model_config.h</code> (<code>QKV</code>, <code>O</code>, <br><code>G</code>/<code>GU</code>, <code>D</code>, with <code>gu_split = IM*2 > 14336</code>) and its three-candidate lookup <br>order from <code>npu_engine_universal.cpp</code>, including the dimension-keyed <br>candidate that <code>xp()</code> returns without checking existence.<br> <li> Reimplements <code>model_tag</code> derivation (stem, parent dir when the stem is <br><code>model</code>, lowercasing, <code>-</code>/<code>.</code>/<code>\</code> normalization, one suffix stripped from <br><code>{_npu2,_instruct,_it,_it_npu2}</code>).<br> <li> Default mode reports both halves per model: missing xclbins (hard <br>stop) and missing <code>insts_i8_*.txt</code> (silent runtime-generator fallback <br>producing wrong results against multi-row v27 xclbins), with <code>--json</code> <br>output and a summary count.<br> <li> <code>--pairing</code> mode audits the committed xclbin set, using quoted literals <br>scraped from the engine source to classify unpaired artifacts as <code>RISK</code> <br>(engine-loaded slot) or a mere note, plus orphan instruction files.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2483/files#diff-cbe1c374199ebd17030f8a5f04ee1b7ed7262728798603bc9f3ee98c52dcda25">+306/-0</a>&nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
## Added: a second commit — the pairing half, and a refuted hypothesis

The PR now carries a second commit adding `--pairing`, because the census answered "does this slot resolve?" but not "did this xclbin ship its instructions?" — and `init_i8()` documents the second one as a **wrong-answer** path, not a slow one: when the slot's `.txt` is absent it falls back to the runtime generator, which emits single-core-row instructions that silently compute the wrong result against a multi-row (v27) xclbin (`npu_engine_universal.cpp:859`).

Measured over `origin/main`'s `engine/npu/xclbins`: **122 `final_i8_*.xclbin` against 120 `insts_i8_*.txt`**, so two ship without instructions:

```
final_i8_ATTN_qwen3_0_6b.xclbin
final_i8_ATTN_v.xclbin
```

Neither is a live risk — the engine's slot vocabulary has **no `ATTN` literal** (it loads `QKV/O/G/U/GU/D/GUSILU_i4` plus `MOE_GU/MOE_D/MOE_SGU/MOE_SD/MOE_GUSGU/MOE_DSD`), so those belong to another loader. The tool decides that per artifact rather than from a hardcoded exception list, by finding the longest underscore prefix of the artifact name that the engine names as a quoted literal.

**Mutation-checked**, so it flags real fallbacks and not just unrecognised names — removing `insts_i8_MOE_GU_zaya.txt` yields:

```
RISK   final_i8_MOE_GU_zaya.xclbin — slot 'MOE_GU' IS loaded by the engine, so
       this falls back to the runtime generator (wrong results)
```

The committed set reports **0** of those.

### It also refutes a hypothesis about #2307

I went looking there first, because "missing insts → the generator → garbage" was the cheapest explanation for that issue's `rms=4.6e8 / corr=-nan`. It is wrong: `insts_i8_MOE_GUSILU_i4_zaya.txt` **is** committed, so the non-fused int4 MoE path has its instruction file. Worth recording so the next person does not spend a round on it.

The default mode is unchanged — 18 models, 85 slots, 33 without an xclbin, the same numbers as when this tool landed.

bong-water-water-bong:
## Third commit: the deployed-model census now resolves both halves

The tool resolved `xp()` per model and slot but not `ip()`, and those are not the same kind of failure:

| condition | what the engine does |
|---|---|
| no xclbin | `init_i8()` returns false, the caller prints `FAIL <slot>` and exits — **loud** |
| no `.txt` | drops to the runtime generator, which emits single-core-row instructions that, against a multi-row (v27) xclbin, compute the **wrong answer** — **silent** |

So "an xclbin with no instruction file" is the row worth counting, and nothing reported it. The tool now resolves both halves per slot, marks that combination `<- RUNTIME GENERATOR (wrong results)`, and counts it separately from the no-xclbin total.

**Measured over 18 deployed models × 85 slots: 33 slots with no xclbin (unchanged) and 0 in the dangerous state.** So this lands as a guard, like `--pairing`, not as a live finding.

**Mutation-checked:** removing `insts_i8_QKV_qwen3_0_6b.txt` makes Qwen3-0.6B report

```
QKV K=1024 N=4096 via tag final_i8_QKV_qwen3_0_6b.xclbin  insts MISSING
    <- RUNTIME GENERATOR (wrong results)
```

with `generator_fallback_slots: 1` in `--json` and the model and slot named; restoring the file returns it to 0.

This completes the answer the tool gives for a deployed model: *does this slot have an xclbin* (round 60), *did the committed xclbin ship its instructions* (`--pairing`), and *will this model silently take the generator* (this commit).

PR #2444, opened 2026-09-17
census: map picolm -> RCPP_ARCH_PICO (auto-draft)

bong-water-water-bong:
## Auto-filed by the census watcher

The daily new-model watcher found an uncovered HF architecture (`picolm`) that looks like a known family. This draft proposes the one-line alias so it stops surfacing as a coverage gap.

**Proposed mapping:** `picolm` → `RCPP_ARCH_PICO` (candidate, needs human review)

**Models triggering it:** aethertp/PicoLM-80M-Instruct

**Steps for the reviewer:**
1. Confirm `picolm` really is the `RCPP_ARCH_PICO` family — on the checkpoint's tensors, not on the class name or the config. A config can match a family exactly and still be a different model (see `vapor` in `Testing/arch-gaps.md`). `model.safetensors.index.json` answers it in one request, or the safetensors header by range request when the model is a single shard.
2. If yes, mark ready + merge. If it's a *new* architecture, close this and record the evidence in `Testing/arch-gaps.md` under *Uncovered classes reviewed later* instead — a class that is neither a rename nor a shape match is not a mapping problem.

Auto-generated by `Testing/census_autopr.py` — the mapping is a *guess*, not verified.

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Map census alias `picolm` to `RCPP_ARCH_PICO`

- Commit Qwen3-4B split-G xclbins (G, U, D) and insts

- Add `build_q34b_split_g.sh` recipe with aiecc-arm and cols findings

- Regenerate `PROVENANCE.json`; drop GitNexus counts in docs

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["census gap: picolm"] -- "alias added" --> B["RCPP_ARCH_PICO in bitnet_model.h"]
  C["build_q34b_split_g.sh"] -- "generates" --> D["G/U_K2560_N9728 + D_K9728_N2560 xclbins"]
  D -- "recorded in" --> E["PROVENANCE.json, 89 builds"]
  D -- "fixes FAIL G" --> F["npu_engine_qwen3_4b init"]
  G["AGENTS.md / CLAUDE.md"] -- "index stats removed" --> H["GitNexus header"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>2 files</summary><table>
<tr>
[...]

PR #2500, opened 2026-09-17
fix(census): only file an alias PR for an exact name, and never re-file a reviewed class

bong-water-water-bong:
## What

`Testing/census_autopr.py` opens a draft PR when the daily watcher finds an uncovered HF architecture class. Its candidate gate was **name similarity alone**, so it filed a PR for any name that merely *looked* like a mapped one. Every PR it has ever filed was closed unjustified:

| PR | proposed | why it was wrong |
| --- | --- | --- |
| #2443 | `language` → `RCPP_ARCH_OBILANGUAGE` | `difflib` ratio 0.842 |
| #2444 | `picolm` → `RCPP_ARCH_PICO` | `"picolm".startswith("pico")` |

That is 2 filed, 2 closed, 0 merged — and the cause is structural, not two unlucky inputs. A third guess, `lladamodellm`, got as far as a pushed branch and no PR at all.

### The same wildcard, in the production log

This is not only a synthetic-corpus result. The workflow's own logs name the rule that fired, for candidates it found on real arrivals:

```
2026-09-13  [autopr] haiku   -> RCPP_ARCH_LLAMA (candidate from h)
2026-09-14  [autopr] han2han -> RCPP_ARCH_LLAMA (candidate from h)
```

`(candidate from h)` is the one-character alias. Neither became a PR — but only because those two runs' commits failed for an unrelated reason (no git identity in the job), not because the gate rejected them.

The 09-17 candidates did escape: `language` and `picolm` were staged, the commit failed the same way, and the leftover staging was then swept into an unrelated blog-post PR by a later step's bare `git commit` — so both aliases sat in #2450's diff. They have been removed there, and the workflow defect that let them through is filed separately.

### The third case, recorded in the header itself

The fuzzy rule also picked `lladamodellm`, whose branch still sits on origin as `census/auto-map-lladamodellm` — pushed, **no PR ever opened** (`d88c77961`, "auto-propose lladamodellm -> RCPP_ARCH_LLADA2"). The header documents how that ended, at `bitnet_model.h:2569`:

> the alias-autopr guessed LLADA2 from llada2moemodellm, but the live config declares model_type llada, which is already mapped to the config-verified llama profile

The old rule scored `lladamodellm` vs `llada2moemodellm` at ratio **0.857** — over the 0.8 gate — at an **edit distance of 5**, which the corrected `<=2 edits` rule rejects. So the ratio threshold, not any family relation, is what produced that token too. (`RCPP_ARCH_LLADA2` is a real token, `= 64`, so at least it was not a build break.)

## Why it over-fires

The table it matches against (`_known_mappings`, parsed from `include/rocm_cpp/bitnet_model.h`) is an **exact-match dispatch table**, not a family list. It legitimately holds 1–4 char aliases, because some HF `model_type`s really are that short:

```
bitnet_model.h:1213   if (strcmp(s, "h") == 0) return RCPP_ARCH_LLAMA;  // config-verified llama profile (h_model)
bitnet_model.h:637    if (strcmp(s, "rw") == 0) return RCPP_ARCH_FALCON;
bitnet_model.h:931    if (strcmp(s, "h3") == 0) return RCPP_ARCH_LLAMA;
```

Two rules then read that table as families:

* **rule 3** treated every key as a family **prefix**, so the shortest alias became a wildcard — any uncovered class starting with `h` resolved to `RCPP_ARCH_LLAMA`.
* **rule 2**'s comment always said `<=2 substitutions/dels, normalized`, but the code computed `difflib.SequenceMatcher(...).ratio() > 0.8` — **it never implemented the rule it documented**. That is what matched `language` vs `obilanguage` (ratio 0.842, but edit distance 3).

Measured against `origin/main`, same header and same inputs:

```
BEFORE  would file: ['kimi.k3', 'language', 'picolm']
AFTER   would file: ['kimi.k3']            alert only: ['picolm']
```

and, over probe corpora:

| probe set | produced a candidate (before) |
| --- | --- |
| 38 real non-family HF arch/model names | **24 (63%)** |
| 300 uniform-random 6–12 char tokens | 12 (4%) — all 11 `h`-initial tokens, plus one `rw`-initial; i.e. every token beginning with a 1–2 char alias |
| keys in the table with normalized length ≤ 4 | 175 |

No fuzzy rule has a **demonstrated true positive**: both examples in the module docstring are already aliases in the header, and rule 1 (exact normalized match) is what reaches them —

```
bitnet_model.h:1601   if (strcmp(s, "kimi_k3")   == 0) return RCPP_ARCH_KIMI_K3;
bitnet_model.h:1602   if (strcmp(s, "kimik3")    == 0) return RCPP_ARCH_KIMI_K3;  // HF model_type (no underscore)
bitnet_model.h:1608   if (strcmp(s, "qwen35")    == 0) return RCPP_ARCH_QWEN35;
bitnet_model.h:1609   if (strcmp(s, "qwen35moe") == 0) return RCPP_ARCH_QWEN35;
```

so `qwen35moe` can never reach `_guess` at all.

The drafter already knew a name is not evidence — the PR body it writes says so:

> Confirm `{arch}` really is the `{target}` family — on the checkpoint's tensors, **not on the class name or the config**.

It selected on exactly the signal it told the reviewer to distrust.

## Changes

* **Only an exact normalized-name match files.** That is a spelling difference (`kimik3` vs `kimi_k3`), not a claim about the architecture. This path is unchanged.
* **Fuzzy matches are alert-only.** The candidate is printed for a human; no PR is opened. `CENSUS_AUTOPR_FUZZY=1` restores the old behaviour for anyone who wants it.
* **Rule 2 now computes Levenshtein distance** with a length floor, as its comment always claimed. **Rule 3** requires the key to be ≥ 4 chars, so a short alias can no longer be a wildcard.
* **`_known_mappings` takes the FIRST match** for a duplicated alias, mirroring `rcpp_arch_from_string`'s linear `if` chain. It took the last, and the two disagreed in a way I did not expect to find: `qwen3_5moe` is mapped at **two points in that same function** —
  ```
  :1614   if (strcmp(s, "qwen3_5moe")  == 0) return RCPP_ARCH_QWEN35;   // Qwen3.5-MoE
  :2718   if (strcmp(s, "qwen3_5moe")  == 0) return RCPP_ARCH_QWEN3;
  ```
  — so the engine answers `QWEN35` while this tool reported `QWEN3`. It is one of 8 aliases defined twice (the other 7 repeat the same token; `qwen3_5_moe` at `:2719` also goes to `QWEN3`, so the intent there is ambiguous). I have left the header alone — flagging it rather than changing engine behaviour in a tooling PR.
* **`_guess` is deleted, not kept as a shim.** It returned a candidate with no kind, so any caller could re-create this bug by accident. The public entry point (`maybe_file_draft_pr`) keeps its signature.

## Guard

`census_autopr.py --self-test`, wired into `Testing/run_all.sh` next to the other census checks, with `census_autopr.py` added to `census_scripts_selfcheck.py`'s family (it writes to the engine header, so the pinned-`ROOT` rule applies to it).

It pins the two over-fires, the exact-match path that must still file, and table-vs-engine agreement — and it **fails when it can read no table**, so "could not determine" cannot print as a pass. Each of the three guards was verified to fail when the defect it covers is re-introduced.

One note on that guard, because it is the same mistake this PR fixes: **the first version of the wildcard check was itself vacuous.** It skipped keys using `_MIN_FAMILY_LEN`, so setting that constant back to 1 to re-introduce the wildcard also switched the check off — it reported PASS while broken. The bound is now a literal, with the reason in a comment. A guard must not be parameterised by what it guards.

## And it will not re-propose a class that has already been reviewed

The gate above stops a *name* from filing. It does not stop a **settled question**. `Testing/arch-gaps.md` is where a class is recorded once a review concludes it is not an alias — 15 such headings on `main`, `haiku` and `han2han` among them — and nothing read it. So `language` and `picolm` were filed, closed unjustified, and **re-filed by the run that had already found them that morning**.

`maybe_file_draft_pr` now skips a class named in an `arch-gaps.md` heading, before it looks at the candidate kind: a review outranks a name match, and even an exact-looking rename has been looked at already. A heading can name two classes (`### `blockmtp` / `looped_block_mtp``), so every backticked token on a heading line is matched, normalized.

This PR also records the two classes that prompted it, to the standard the other entries in that file use:

* **`language` / `helloboy91/tiny-english-30m`** — `architectures: ["LanguageModel"]` with **no `model_type`**, so `strip_arch()` leaves the bare word `language`. 58 F32 tensors named `blocks.N.*`, with a fused `attention.qkv.weight`. `RCPP_ARCH_OBILANGUAGE` (832) has one alias, no branch in `src/model_router.cpp` and no implementation, so mapping the generic word `language` would silently claim OBILANGUAGE for *any* custom checkpoint that strips down to it.
* **`picolm` / `aethertp/PicoLM-80M-Instruct`** — its own `modeling_picolm.py` rotates **half-split** (`x1, x2 = x[..., :D//2], x[..., D//2:]`, then `torch.cat([x1*cos - x2*sin, x1*sin + x2*cos])`) while `RCPP_ARCH_PICO` is documented as **adjacent-pair** RoPE (`view_as_complex`) — different results for the same weights — and it adds **QK-norm** (`q_norm`/`k_norm`, `[64]`) that the PICO token's description does not mention. `view_as_complex` appears in the tree only inside that enum comment.

Neither belongs in `NON_TEXT_GEN`: both are genuine causal text LMs, so excluding them as "not text-gen" would be a false exclusion. They stay uncovered — the honest state.

The self-test asserts this twice over, because the first version of the check was **vacuous**: it used `haiku`, which no longer produces a candidate at all, so "proposes nothing" was true for the wrong reason. It now requires `_guess_ex("picolm")` to still return a candidate *and* `maybe_file_draft_pr(["picolm", "haiku"])` to propose nothing — a pair only satisfiable if the review record is what does the suppressing. Breaking either half fails the test.

## Verification

* `py_compile` clean on all three touched Python files.
* `census_autopr.py --self-test` → PASS (2009 aliases).
* `census_scripts_selfcheck.py` → PASS, 6 scripts.
* `bash -n Testing/run_all.sh`, and the workflow YAML parses.
* Arch-gaps guard: `maybe_file_draft_pr(["picolm"])` → `[]` with `reviewed in Testing/arch-gaps.md — not an alias, no PR`; breaking the guard, **or** making `picolm` stop being a candidate, each fails `--self-test` (verified both ways).
* `repo_docs_selfcheck.py`, `census_scripts_selfcheck.py`, `seo_claim_selfcheck.py` all pass with the new `arch-gaps.md` entries.
* `_distance` checked against known cases (`kitten`/`sitting` = 3, `language`/`obilanguage` = 3, `picolm`/`pico` = 2) and symmetry over 200 random pairs.
* GitNexus impact on `maybe_file_draft_pr`: **LOW** risk, epistemic *exact*, 2 impacted (`census_autopr.py`, `hf_new_models.py`). `detect-changes` cannot analyse a worktree (the index is bound to the main clone and the CI runner copy), so the graph check ran as impact analysis.

#2443 and #2444 are already closed, so this closes nothing. Comments welcome — in particular on whether the fuzzy path should exist at all, or whether the watcher should corroborate on the checkpoint's tensor names before proposing a mapping.

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Stop filing alias PRs on name similarity alone

- Exact normalized-name match still files; fuzzy is alert-only

- Never re-propose a class reviewed in `Testing/arch-gaps.md`

- Parse header first-match-wins; keep hyphenated aliases

- Add `census_autopr.py --self-test` gate to `run_all.sh`

- Refuse short-alias wildcards (`h` -> every h class)

- Document `language` and `picolm` rejections in arch-gaps

- Update autopr workflow and watcher comments

___

### Diagram Walkthrough

```mermaid
flowchart LR
  WATCH["hf_new_models.py watcher"] -- "uncovered archs" --> GUESS["census_autopr._guess_ex"]
  GUESS -- "exact normalized name" --> PR["Draft PR with one-line alias"]
  GUESS -- "fuzzy (name-similar)" --> ALERT["ALERT ONLY, no PR"]
  GAPS["Testing/arch-gaps.md headings"] -- "reviewed class" --> SKIP["suppressed, no PR"]
  GUESS -- "table + engine" --> HDR["bitnet_model.h rcpp_arch_from_string"]
  SELFTEST["census_autopr.py --self-test"] -- "run_all.sh gate" --> GUESS
```

<tr>
  <td>
    <details>
      <summary><strong>census_autopr.py</strong><dd><code>Make alias autopr exact-match only with self-test</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/census_autopr.py

<ul><li>Rewrites docstring: name similarity is not family evidence; fuzzy <br>rules alert-only<br> <li> Adds <code>_guess_ex</code> returning kind <code>exact</code>/<code>fuzzy</code>, with <code>_distance</code> Levenshtein<br> <li> Adds <code>_MIN_FAMILY_LEN</code> / <code>_MIN_FUZZY_LEN</code> floors stopping short-alias <br>wildcards<br> <li> Parses header first-match-wins and allows <code>-</code> in alias keys<br> <li> Adds <code>_documented_in_arch_gaps</code> so reviewed classes are never re-filed<br> <li> Adds <code>_self_test</code> and <code>--self-test</code> CLI, plus <code>CENSUS_AUTOPR_FUZZY</code> escape</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2500/files#diff-0182339e60b9405c8d692fafdda5a68de66470ea1fa3de02be53bc1bd5868e14">+258/-35</a></td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>census_scripts_selfcheck.py</strong><dd><code>Include autopr in census script invariants</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/census_scripts_selfcheck.py

<ul><li>Adds <code>census_autopr.py</code> to the checked script family<br> <li> Enforces the pinned-<code>ROOT</code> rule since it writes to the shared header</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2500/files#diff-0ae0b01ebdf91afa2f3a71ab519da73395e2444c97fec27de4cf4c0a4353f1cb">+3/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>run_all.sh</strong><dd><code>Gate autopr self-test in run_all</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/run_all.sh

<ul><li>Adds a <code>census_autopr --self-test</code> step to the gate suite<br> <li> Fails the run and prints tail output when the self-test breaks</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2500/files#diff-60779953d5d72f0d8ef3aea52f86b0b365bbdb42457aae2e5955d6320e32eb2c">+15/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Documentation</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>hf_new_models.py</strong><dd><code>Clarify alias filing rule in watcher</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/hf_new_models.py

<ul><li>Updates the autopr comment: exact-name aliases file, similar names <br>alert</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2500/files#diff-b2cdcca5305ff83c2aa31abfc2dd476b9458209d9ef12ace923ce85369618ef6">+8/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>census-autopr.yml</strong><dd><code>Document exact-name-only autopr in workflow</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/census-autopr.yml

<ul><li>Rewrites header notes to exact-name-only filing rationale<br> <li> Updates inline step echo and identity comments</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2500/files#diff-0f84ff0173547f1ac1e7e8aad568b593cee71209bc3dd72fb384507984b9efe0">+15/-8</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>arch-gaps.md</strong><dd><code>Record rejected `language` and `picolm` aliases</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/arch-gaps.md

<ul><li>Adds reviewed entries for <code>language</code> and <code>picolm</code> with tensor evidence<br> <li> Records why OBILANGUAGE and PICO aliases are wrong<br> <li> Notes both came from autopr lexical similarity, fixed in #2500</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2500/files#diff-5e896e9c6b325e941db08d8b90ebaba4b8354d83269b195c3ec7562e105e77a7">+73/-1</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Queue prediction from `git merge-tree` against current `main`: **this entry conflicts with `main`**,
on `.github/workflows/census-autopr.yml`, so the queue will build it and then eject it.

The collision is not with your change's logic — `main` gained a git-identity block at exactly the
lines this PR rewrites:

```
              git config user.name "github-actions[bot]"
              git config user.email "github-actions[bot]@users.noreply.github.com"
              # CENSUS_SKIP_PR stays UNSET so census_autopr may file draft PRs.
```

`main` now has those two `git config` lines immediately before the `CENSUS_SKIP_PR` comment that this
PR extends ("for an exact normalized-name match only"), and the merge base has neither, so both sides
edited the same region. The resolution keeps `main`'s identity lines and your extended comment.

Worth knowing for the retry: **#2536 changes this**. Once it lands, the workflow reads back the
outcome and polls `mergeable` before enqueuing, so a PR that conflicts with `main` is skipped with a
warning instead of being enqueued, built, and ejected — the entry stops costing the batch a rebuild.
Until then this one will take a queue cycle to discover what `git merge-tree` can tell you in a
second:

```
git fetch origin main:refs/remotes/origin/main
git merge-tree --write-tree origin/main <your-branch>   # rc 1 = it will be ejected
```

Issue #2505, opened 2026-09-17
ci(npu): no required check compiles the NPU engine with XRT — its source set lives in three places and only bench.yml's hand-rolled list links it


bong-water-water-bong:
## Summary

`npu_engine_universal.cpp` is compiled with XRT by exactly **one** CI job, and that job builds it from a **hand-maintained source list** rather than from the engine's own build description. When the two drift, no *required* check notices: the engine's link breakage is invisible to all four required checks and surfaces only in the non-required `benchmark` job.

Found from PR #2435, which adds three bridge translation units and links nowhere in CI.

## The source set lives in three places

| place | what it lists |
|---|---|
| `engine/npu/CMakeLists.txt` (`NPU_ENGINE_SOURCES`) | `npu_engine_universal.cpp`, `dequant_q4nx.cpp`, `gemm_npu_instructions.cpp`, `zaya_decode.cpp` (+ `npu_runlist_bridge.cpp` on #2435) |
| `engine/npu/build_npu.sh` (`ENGINE_OBJS`) | the same plus every bridge TU the branch adds |
| `.github/workflows/bench.yml` ("Build engine (continuous batch)") | a **hand-written `g++` line**: `npu_engine_universal.cpp`, `build/dequant_q4nx.o`, `zaya_decode.cpp`, `gemm_npu_instructions.cpp`, `npu-infer/src/flm_bridge.cpp` |

Nothing compares them.

## Why no required check catches it

`CMakeLists.txt:2351`:

```cmake
if(XRT_FOUND)
    add_subdirectory(engine/npu EXCLUDE_FROM_ALL)
else()
    message(STATUS "npu: XRT NOT found — skipping NPU engine build (npu_engine_universal will not be a target)")
endif()
```

The required check `C++ (cmake configure + build)` runs on a GitHub-hosted runner without XRT — `ci.yml:38` says so itself ("libxrt-dev not available — fused NPU backends stubbed"). So on that runner `engine/npu` is **not added at all**, `npu_engine_universal` is not a target, and `npu_engine_universal.cpp` is never compiled. The required check cannot fail on an NPU-engine link error by construction.

That leaves `bench.yml`'s `g++` line as the only XRT-enabled compilation of the engine in CI — and it is a copy, not the source of truth.

## The proof, from #2435

[...]

PR #2507, opened 2026-09-17
ci(bench): the benchmark job's trigger did not cover what it compiles

bong-water-water-bong:
## The defect

`bench.yml` runs on `push` filtered to a narrow set of paths — but the job's own compile line reads more than those:

```
sources compiled : engine/npu/src/{npu_engine_universal,gemm_npu_instructions,zaya_decode}.cpp,
                   engine/npu/build/dequant_q4nx.o, npu-infer/src/flm_bridge.cpp
include dirs     : engine/npu, engine/npu/include, engine/npu/generators,
                   include, npu-infer/include
paths: filter    : .github/workflows/bench.yml, engine/npu/src/**, engine/npu/kernel/**,
                   engine/npu/xclbins/**
```

A change to `engine/npu/include/**`, `include/**`, `npu-infer/**`, or to `engine/npu/CMakeLists.txt` — where the job's source list actually lives — **could not fire the job at all**. A gate that silently never runs, which is check 4's own stated theme one step in: the trigger paths exist, they just do not subscribe to what the job builds.

This matters more than a normal path-filter gap because of #2505: `bench.yml` is the **only** job that compiles the engine with XRT. The required `C++ (cmake configure + build)` runs without XRT, so `CMakeLists.txt:2351` adds `engine/npu` only `if(XRT_FOUND)` and never builds the target. This filter is the sole trigger for the sole engine build in CI.

## The fix

```
- 'engine/npu/**'     # covers src, kernel, xclbins, include, generators, CMakeLists.txt, build_npu.sh
- 'npu-infer/**'      # flm_bridge.cpp (compiled) + npu-infer/include (an -I)
- 'include/**'        # an -I, and the engine's own headers land here
```

## The guard

Check 4 in `Testing/repo_docs_selfcheck.py` now asserts the invariant for any workflow that has **both** a `paths:` block and a compiler command: every in-repo source and every `-I` directory that compiler reads must be covered by the filter. It derives the inputs from the workflow's own compile line, so adding `-Ifoo/bar` later either gets covered or fails the check.

Verified on the real file at each step:

| state | result |
|---|---|
| before the fix | **6 findings** — `engine/npu`, `engine/npu/generators`, `engine/npu/include`, `include`, `npu-infer/include`, `npu-infer/src/flm_bridge.cpp` |
| after the fix | **0 findings** |
| drop `npu-infer/**` | **2 findings** — exactly the npu-infer inputs |
| restore the original filter | **6 findings** — reproduces the before case exactly |

## Two mistakes this cost, and why they are in the commit message

Both are this check's own subject matter, so they are worth naming rather than quietly fixing:

1. The first version matched compiler invocations with `\b(g\+\+|gcc)\b`. **`\b` can never match after `g++`** — `+` is not a word character — so it found no inputs in this workflow and reported every workflow clean.
2. The first `paths:` reader stopped at the first line that was not a list item, so the **comment I wrote inside that block** ended the parse and every glob after it read as "not covered" (10 findings, including inputs the globs did cover).

Both now have guards: no `\b` in the invocation regex, comments/blanks skipped by the block reader, and **a job that runs a compiler yet parses no in-repo input is itself a finding** — so the vacuous-pass failure mode cannot come back silently.

`engine/npu/**` also has the side effect of firing this job when `CMakeLists.txt` or `build_npu.sh` change, which is exactly the drift #2505 is about.

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Widen `bench.yml` `paths:` trigger to `engine/npu/**`

- Add `npu-infer/**` and `include/**` to trigger

- Add paths-coverage rule to `Testing/repo_docs_selfcheck.py`

- Rule parses `g++`/`clang` and `cmake -S` inputs

- Handle comments, continuations, `-I` dirs, `dir/**` globs

- Relocate rule anchor to avoid merge-queue conflicts

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["bench.yml paths: trigger"] -- "widened to" --> B["engine/npu/** , npu-infer/** , include/**"]
  C["selfcheck check 4"] -- "scans" --> D["run: blocks"]
  D -- "g++ / cmake -S" --> E["sources and -I directories"]
  E -- "compared to" --> A
  E -- "not covered" --> F["finding: build input not covered"]
  D -- "no inputs parsed" --> G["finding: vacuous check"]
```

<tr>
  <td>
    <details>
      <summary><strong>repo_docs_selfcheck.py</strong><dd><code>Add workflow paths-coverage rule for build inputs</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/repo_docs_selfcheck.py

<ul><li>Add new helpers to detect build inputs: <code>CC_INVOCATION</code>, <br><code>CMAKE_CONFIGURE</code>, <code>strip_comments</code>, <code>run_scripts</code>, <code>trigger_paths</code>, <br><code>compiled_inputs</code>, <code>cmake_inputs</code>, <code>path_covered</code>.<br> <li> Extend check 4 so a job with a <code>paths:</code> block must cover every in-repo <br>source, <code>-I</code> directory and <code>cmake -S</code> <code>CMakeLists.txt</code> its own compiler <br>commands read.<br> <li> Report a finding when a build runs but no in-repo input was parsed <br>(vacuous-check guard), and update the module docstring for check 4.<br> <li> Move the rule to the top of the per-workflow loop body so it does not <br>collide with the <code>#2503</code> conflict-marker check insertion.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2507/files#diff-952d19d481679939123604164fbd207630e4ff66342c79cec423172ddd5e5230">+175/-1</a>&nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Configuration changes</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>bench.yml</strong><dd><code>Widen benchmark push trigger to all build inputs</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/bench.yml

<ul><li>Replace the three narrow <code>push</code> path entries with <code>engine/npu/**</code>, <br>covering <code>src</code>, <code>kernel</code>, <code>xclbins</code>, <code>include</code>, <code>generators</code>, <code>CMakeLists.txt</code> and <br><code>build_npu.sh</code>.<br> <li> Add <code>npu-infer/**</code> (compiled <code>flm_bridge.cpp</code> plus its include dir) and <br><code>include/**</code> so header changes re-run the job.<br> <li> Document that this is the only XRT-enabled engine build in CI and that <br>the list is checked by <code>Testing/repo_docs_selfcheck.py</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2507/files#diff-a5e740be96415373789689f814583e93ff2a8f05eae6481e94505fd6cb6bc6a7">+10/-3</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
# A concrete instance of the cost, from a commit of mine

The aggregate numbers are further down; here is one case, because it is the kind of thing that is easy to
read past. I pushed a commit that changed **only comments** in `bench.yml`:

```
9eef07b0f docs(bench): the artifacts were never missing — it was the search path
  .github/workflows/bench.yml | 22 ++++++++++++++++------
  1 file changed, 16 insertions(+), 6 deletions(-)      <- all of it inside `#` comments
```

and it fired a full benchmark run:

```
03:17:27 queued  run=35302646313
```

because `paths:` lists `.github/workflows/bench.yml` as one of its own inputs. That is correct in itself —
the workflow *is* a dependency of the job it describes, and a change to the build step must re-run it —
but it means every comment edit in this file costs an engine build plus an NPU benchmark on the one box
that has an NPU, which is the same box seven `End-to-End Smoke Test` runs were queued behind at the time.

I cancelled that particular run (it could only have measured the same binary as the previous one), which
is a one-off and not a fix. The point for this PR is narrower: the firing rate is the *product* of the
paths coverage and the ref coverage, and this PR widens the first while the second stays at "every branch
and every merge-queue ref". `branches: [main]` on the same `push:` block would not have prevented this
case, but it is what takes the volume from ~42 runs/6h to ~6.

bong-water-water-bong:
Independent verification of this PR's claim after it merged, by driving the cited
rule directly rather than trusting the sentence.

`Testing/repo_docs_selfcheck.py` exposes the rule as pure functions
(`trigger_paths`, `run_scripts`, `strip_comments`, `compiled_inputs`,
`cmake_inputs`, `path_covered`), so both states can be fed to it as text:

    pre  (091699210) paths = ['.github/workflows/bench.yml', 'engine/npu/src/**',
                              'engine/npu/kernel/**', 'engine/npu/xclbins/**']
         job inputs detected = ['engine/npu/CMakeLists.txt']
         UNCOVERED           = ['engine/npu/CMakeLists.txt']

    post (origin/main) paths = ['.github/workflows/bench.yml', 'engine/npu/**',
                                'npu-infer/**', 'include/**']
         job inputs detected = ['engine/npu/CMakeLists.txt']
         UNCOVERED           = none

So the rule **can** fail — it fires on the known-bad state this PR fixed and reports
nothing after it — which is the check the PR body asserted and nobody had run
against the pre-state. `engine/npu/CMakeLists.txt` is the interesting one: it is
where `NPU_ENGINE_SOURCES` lives, i.e. exactly the file #2505 is about, and it was
the uncovered input before this PR.

Also ran a wider sweep of the same class over every workflow with a
`paths:`/`paths-ignore:` block (14 of 34), comparing trigger globs against the files
the workflow's own run bodies reference. Apparent hits all resolved:
`site/benchmarks.js` twice (a prefix artifact of `benchmarks.json` — that file does
not exist), `scripts/generate_benchmark_table.py` (guarded as retired, with an
explicit notice), `Testing/arch-gaps.md` (a write target, not an input),
`tools/validate_claims.py` (covered — `Testing/run_all.sh` runs it), and the
remaining `scripts/*.py` cases, which are triggered by sibling workflows or by
content changes. No second instance of this defect class was found; that is a
negative result and it is only meaningful because the rule above was first shown to
be able to fail.

PR #2515, opened 2026-09-17
fix(npu): the GDN fused-qkv pack advanced its source pointer by rows, not by rows*H (#2451)

bong-water-water-bong:
### **User description**
Part of #2451 — the GDN half only; #2451 stays open for the STD restructure.

## The defect, verified from source rather than taken from the report

`transpose_pack` (engine/npu/src/npu_engine_universal.cpp:467) is

```c
dst[i * dst_stride + dst_offset + o] = src[o * in_f + i];    // o < out_f, i < in_f
```

so `src` is the head of a row-major `[out_f][in_f]` block and **`in_f` is that block's row
stride**. The GDN branch passes `in_f = H`, and `gdn_k_off` / `gdn_v_off` are *row* counts —
the same values it passes as `out_f` and as `dst_offset`:

```c
int gdn_k_off = (gdn_vh[l] / 2) * gdn_hd[l];        // rows of q  (q is (vh/2)×hd)
int gdn_v_off = gdn_vh[l] * gdn_hd[l];             // rows of q + rows of k
transpose_pack(qkv_w,                       gdn_k_off, H, ...);   // Q   - row 0, no offset
transpose_pack(qkv_w + gdn_k_off,           gdn_k_off, H, ...);   // K   <- rows, not rows*H
transpose_pack(qkv_w + gdn_v_off,           gdn_v_off, H, ...);   // V   <- rows, not rows*H
```

Reproduced on the index arithmetic alone (H=8, `gdn_k_off`=4), without a device:

```
  K block rows selected by qkv_w + gdn_k_off      (old): [0, 1, 2, 3, 4]
  K block rows selected by qkv_w + gdn_k_off * H  (fixed): [4, 5, 6, 7]
  intended: [4, 5, 6, 7]
```

Only row 0 is right by accident — head 0 — which is what makes this read as per-head noise
rather than as a whole block read from the wrong place. The same asymmetry is why a "GDN layers
look fine" inference is unsound: `origin/family/head-block-loop:1880-1881` fixes both GDN sites
with exactly `(size_t)gdn_k_off * H` / `(size_t)gdn_v_off * H` — reachable, an independent
derivation — and the unpushed `~/tip-wt` line carries the same form.

## Not in this PR

The STD per-head sites (`:1489-1490`). `* H` is the obvious drive-by fix there and it is
**wrong**: the converter flattens full-attention q/gate into `[q_all | gate_all]`
(`third_party/FLM_Q4NX_Converter/q4nx/models/qwen35.py`: `rearrange(..., '(g p h) c -> (p g h) c', p=2, h=DH)`,
fired on `layer_id % full_attntion_interval == interval-1`), so that branch needs the
restructure — plus the QKV-context widening, or the 8192-row layout still meets a 6144-row
context. That half is measured on-device by the lane that owns it and #2451 remains open for it.

## The guard

`Testing/npu_pack_stride_selfcheck.py`, wired into `Testing/run_all.sh`: a source-level
invariant, no device needed, because the packing lives inside a 4000-line function and a wrong
block still has the right shape — nothing else in the suite can see it.

* every `buf + <expr>` source offset must scale `<expr>` by `H`;
* the two STD sites are named as exemptions *with the restructure reason*, and the check
  **fails if an exemption stops matching a call site** — so landing the restructure trips this
  check until the exemption is deleted, instead of leaving a stale allowance behind;
* floors on the total call count (22 today) and on the offset-call count (4 today) so a scan
  that quietly stopped matching cannot report a clean result.

Verified in both directions: **exit 1** on `origin/main`'s file, **exit 1** with the two GDN
sites re-broken in place, **exit 1** on a stale exemption, **exit 0** here.

```
npu pack stride - 2 file(s), 22 transpose_pack call(s), 4 with a source offset
  exempt (still unscaled): qkv_w + h * 2 * std_hd[l]
  exempt (still unscaled): qkv_w + h * 2 * std_hd[l] + std_hd[l]
OK: every source offset is scaled by H (2 scaled, 2 exempt)
```

`cmake --build --target npu_engine_universal` → 16.8 s, 1,334,288-byte binary. `bash
Testing/run_all.sh` → `27/41 passed, 14 skipped`, `✓ npu_pack_stride`.

Graph analysis: GitNexus reports 3 candidates for `transpose_pack` with `risk: UNKNOWN` — both
definitions are `static` (file-local), and a text search of every call site (14 in the universal
engine, 5 in the cb worker, all listed above) confirms the change alters only the argument
expression at two call sites; the function's signature and semantics are untouched.

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Scale GDN fused-qkv K/V pack offsets by `H`

- Source pointer now advances `rows * H`, not `rows`

- Add source-level stride self-check with explicit exemptions

- Deliberately leave STD per-head sites unfixed (needs restructure)

___

### Diagram Walkthrough

```mermaid
flowchart LR
    A["GDN K/V pack: qkv_w + gdn_k_off"] -- "offset was rows, not rows*H" --> B["K/V read from window inside q's first rows"]
    B -- "fix" --> C["qkv_w + (size_t)gdn_k_off * H"]
    C -- "guarded by" --> D["Testing/npu_pack_stride_selfcheck.py"]
    D -- "wired into" --> E["Testing/run_all.sh"]
    F["STD per-head q/gate sites"] -- "exempted: needs [q_all | gate_all] restructure" --> D
```

<tr>
  <td>
    <details>
      <summary><strong>npu_engine_universal.cpp</strong><dd><code>Scale GDN fused-qkv K/V source offsets by row stride H</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

<ul><li>Fix the GDN fused-QKV K block source pointer to <code>qkv_w + </code><br><code>(size_t)gdn_k_off * H</code> instead of <code>qkv_w + gdn_k_off</code>.<br> <li> Fix the V block the same way (<code>qkv_w + (size_t)gdn_v_off * H</code>); the Q <br>block keeps its zero offset.<br> <li> Add a six-line comment stating that <code>transpose_pack</code> reads <code>src[o * in_f </code><br><code>+ i]</code> with <code>in_f == H</code> while <code>gdn_k_off</code>/<code>gdn_v_off</code> are row counts (also <br>used as <code>out_f</code> and <code>dst_offset</code>), and pointing at the new regression <br>guard.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2515/files#diff-63677a636d5ff19712cf034cc1e2c15a96b194137caee54d8bd6883b0b76a10b">+8/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>npu_pack_stride_selfcheck.py</strong><dd><code>Add stride self-check for fused-weight transpose_pack offsets</code></dd></summary>
<hr>

Testing/npu_pack_stride_selfcheck.py

<ul><li>New source-level self-check that scans every <code>transpose_pack</code> call in <br>the two NPU engine files and inspects the first argument.<br> <li> Flags any offset call of the form <code>buf + <expr></code> whose expression is not <br>scaled by <code>H</code>, using the regex <code>\*\s*H\b</code>.<br> <li> Carries a <code>KNOWN_UNSCALED</code> exemption map for the two STD per-head q/gate <br>sites, with the reason (converter <code>(g p h) -> (p g h)</code> restructure), and <br>fails if an exemption no longer matches a call site.<br> <li> Adds <code>MIN_CALLS</code> / <code>MIN_OFFSET_CALLS</code> floors so a scan that silently stops <br>matching cannot pass, plus exit codes 0/1/2 for <br>ok/violation/environment error.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2515/files#diff-2052e8a960be39917f914672a23b31d0fbb5e986efec81693419ee1fdf9e6d22">+174/-0</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>run_all.sh</strong><dd><code>Wire the NPU pack stride self-check into run_all</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/run_all.sh

<ul><li>Register the new check as an <code>npu pack stride</code> suite step, incrementing <br><code>total</code> and <code>fail</code> like the other cases.<br> <li> Run <code>$PYTHON Testing/npu_pack_stride_selfcheck.py</code>, indenting full <br>output on success and the last 6 lines on failure.<br> <li> Add a comment explaining the defect class (unscaled row block inside a <br>4000-line function, invisible to other checks).</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2515/files#diff-60779953d5d72f0d8ef3aea52f86b0b365bbdb42457aae2e5955d6320e32eb2c">+16/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Coordination note: #2479 lands the **same GDN hunk** as this PR (byte-identical) plus an STD change that takes the `* H` offset form. Whichever merges second should drop its GDN half.

This PR's guard (`Testing/npu_pack_stride_selfcheck.py`) exempts the STD sites on purpose — its comment says the correct fix there is the restructure, not an offset — so if #2479 lands as-is the guard fails on the stale exemption, by design, and I will rebase this PR to match whatever the row question settles to. See my comment on #2479 for why the converter's `(g p h) -> (p g h)` rearrange plus the lane's `corr(q)` measurement point at the restructure.

Issue #2459, opened 2026-09-17
investigation(npu): the running amdxdna is a deleted legacy tree predating upstream BO-teardown/mailbox fixes — P0.1 status may be stale


bong-water-water-bong:
## Summary

This box runs an `amdxdna` built from the **legacy out-of-tree tree that AMD deleted upstream on 2026-09-01**, and it predates three fixes landed since in the BO-teardown / MMU-notifier / mailbox paths — the subsystem behind the chronic exec-fault storm tracked as `research/TRACKING.md` P0.1. Moving to the upstream staging driver is now the only supported path and would bring them; nobody appears to have assessed that.

## Evidence

**The running module is the deleted tree.** `/lib/modules/7.2.0-next-20260821-unstable-ogc-g2a559b27-1/kernel/drivers/accel/amdxdna/amdxdna.ko`, built 2026-09-01 11:50, carries the compile-path string:

```
$ strings amdxdna.ko | grep -oE 'src/driver/amdxdna|drivers/accel/amdxdna' | sort -u
src/driver/amdxdna
```

and it taints the kernel on load (`amdxdna: loading out-of-tree module taints kernel`).

**Upstream removed that tree.** `813e0bf` — Max Zhen, 2026-09-01:

> **Remove out-of-tree (legacy) driver under src/driver/**
> The upstream driver in `drivers/accel/amdxdna/` is the sole driver going forward. Remove the legacy `src/driver/` tree and all references to it, including the `PACKAGE_LEGACY_DRIVER` CMake option…

**Three landings postdate the running build, in the affected subsystem** (`amd/xdna-driver`, 41 commits between 2026-08-28 and 2026-09-15, linear):

| commit | date | subject |
|---|---|---|
| `ebd297c` | 2026-09-01 | accel/amdxdna: check the ring buffer before failing a mailbox timeout — adds `xdna_mailbox_drain_channel()` |
| `85f9dcb` | 2026-09-09 | accel/amdxdna: fully init client before adding it to client_list |
| `77e5325` | 2026-09-10 | accel/amdxdna: unregister MMU notifier before removing BO from invalid list — adds `amdxdna_hmm_unregister()` + `flush_workqueue(xdna->notifier_wq)` before removal |

## Why this looks relevant to P0.1 — and what I am *not* claiming

[...]

PR #2510, opened 2026-09-17
test(suite): wire the two orphaned backend selfchecks, and trip on the next one

bong-water-water-bong:
## Two selfchecks that nothing ran, and one that had already rotted

A scan of every `Testing/*_selfcheck.*` against `run_all.sh` and every workflow found exactly **two never invoked**:

| selfcheck | documented as | invoked by |
|---|---|---|
| `Testing/hrx_backend_selfcheck.cpp` | *"→ **10/10 passed**"* (`docs/research/hrx-backend.md`) | nothing |
| `Testing/lse_backend_selfcheck.cpp` | *"M0 green: spawns real lse-server…"* (`docs/plans/lse-backend-integration.md`) | nothing |

**The HRX one had rotted.** Its own header documents:

```bash
g++ -std=c++17 -Iinclude -Isrc -Ibuild/_deps/nlohmann_json-src/include \
    src/backend_hrx.cpp Testing/hrx_backend_selfcheck.cpp -o /tmp/hrx_check
```

and `hrx::Inprocess` — the class the check drives — lives in **`src/hrx_inprocess.cpp`**, which that line omits. The link fails on **ten undefined references** (`hrx::Inprocess::{init,generate,load_model,load_session_file,has_hrx_device,hrx_device_name,reset,unload}`, ctor and dtor). Adding the file makes it compile and pass:

```
ok   factory_returns_backend
ok   type_is_HRX_GPU
ok   not_initialized_yet
skipped live spawn (set HRX_ROOT + HRX_TEST_MODEL)
```

Nobody could have noticed: a check nothing runs cannot report that it no longer builds.

## What this does

* Wires both into `run_all.sh`, with the source list each actually needs and a loud skip when the fetched nlohmann include is absent (a tree that was never configured) — the same shape as the build-gated blocks around them.
* Corrects the HRX header's stale build line, which is what would have misled whoever wired it up next.
* **Adds a tripwire:** every `Testing/*_selfcheck.*` must be invoked by `run_all.sh` or a workflow, or the suite fails naming the orphan.

## Verification

```
on this box (build tree present)    ✓ hrx-backend   ✓ lse-backend   28/42 passed, 14 skipped
without a build tree (as in CI)     - hrx-backend: nlohmann include absent, skipped
                                    - lse-backend: nlohmann include absent, skipped
                                                                     27/43 passed, 16 skipped
```

[...]

PR #2518, opened 2026-09-17
fix(npu): the driver-reapply script looked for a filename this kernel does not have, and would have reinstalled the tree upstream deleted (#2517)

### LINKED ISSUE #2517: ops(npu): the driver-reapply safety net cannot run on this box — it looks for amdxdna.ko.zst, and the module it would rebuild comes from the tree upstream deleted
`scripts/npu-driver-reapply.sh` exists so a kernel upgrade does not silently revert the NPU stack
(its header: "Kernel upgrades silently restore the stock in-tree module — run this after every
kernel update"). On this box it **cannot run at all**, and the module it would rebuild comes from
the tree upstream deleted on 2026-09-01 — #2459's finding, with nothing in the script that
notices either fact.

## 1. It looks for a filename that does not exist here

```
$ grep -n 'KO=' scripts/npu-driver-reapply.sh
19:KO="$KO_DIR/amdxdna.ko.zst"
$ bash scripts/npu-driver-reapply.sh
ERROR: /lib/modules/7.2.0-next-20260821-unstable-ogc-g2a559b27-1/kernel/drivers/accel/amdxdna/amdxdna.ko.zst not found — is kernel 7.2.0-next-20260821-unstable-ogc-g2a559b27-1 installed?
$ echo $?
1
$ modinfo amdxdna | grep '^filename'
filename:       /lib/modules/7.2.0-next-20260821-unstable-ogc-g2a559b27-1/kernel/drivers/accel/amdxdna/amdxdna.ko
```

The running kernel ships the module **uncompressed**; the script's first check therefore fails
before it can do anything, and the safety net it promises is not there.

## 2. Two more premises that do not hold

* `BUILD="$SRC/build/Release"` — `~/xdna-driver` has `build/` and nothing under it
  (`ls -d ~/xdna-driver/build*` → `build`), and no `amdxdna.ko` exists anywhere in the tree
  (`find ~/xdna-driver -name amdxdna.ko` → nothing). The script would stop at line 44 with
  "build dir … missing — configure first".
* Signing: `SIGN_KEY=/var/lib/dkms/mok.key` does not exist (neither does `mok.pub`), while
  `modinfo amdxdna` reports **no** `signer`/`sig_id` — the module that is loaded and working on
  this box is unsigned. The header's "signed with the local MOK key … SecureBoot + lockdown
  enforce module signatures" is not the state this box is in, so the sign-file step is a third
  way for the script to fail rather than a step that protects anything.

## 3. The provenance problem (this is the part that matters)

The module the script would rebuild comes from the **legacy out-of-tree tree**:

```
$ git -C ~/xdna-driver log -1 --format='%h %ad %s' --date=short
7004f1c 2026-08-27 shim: allow firmware log and event trace queries on mgmt PF
$ git -C ~/xdna-driver merge-base --is-ancestor 813e0bf HEAD || echo "predates the deletion"
predates the deletion                      # 813e0bf, 2026-09-01: "Remove out-of-tree (legacy) driver under src/driver/"
$ git -C ~/xdna-driver status --porcelain | head -3
 M src/driver/amdxdna/aie2_ctx.c
 M src/driver/amdxdna/aie2_pci.h
 M src/driver/amdxdna/amdxdna_gem.c
```

and the installed module is from that same tree — identified the way #2459 identified it, by the
compile path baked into the object:

```
$ strings /lib/modules/$(uname -r)/kernel/drivers/accel/amdxdna/amdxdna.ko \
    | grep -oE 'src/driver/amdxdna|drivers/accel/amdxdna' | sort | uniq -c
      2 src/driver/amdxdna
$ stat -c '%y' …/amdxdna.ko
2026-09-01 11:50:12
```

So the one script whose job is to restore the custom driver would restore the driver AMD removed,
and nothing in it records which tree a rebuild came from. #2459's scope note ("moving to the
upstream staging driver is now the only supported path") is the other half of this; the script is
where it becomes a host fact rather than an observation.

## Fix (PR follows)

Discovery and provenance, no new policy: find the installed module under either compression; read
the tree a checkout and a built module came from; **refuse** the legacy tree and refuse to guess
(exit 2) when the provenance cannot be determined; name the missing MOK key files instead of
failing inside `sign-file`; and correct the header to the state of this box. Nothing here decides
which driver this project should ship — #2459 owns that — it only stops the script from installing
the obsolete one while claiming to be a safety net.

Not run by CI or by any automation (only `docs/CODEBASE.md` lists it), so this is an ops-script
defect, not a gate defect.

bong-water-water-bong:
### **User description**
Fixes #2517. Related: #2459 (which driver this project should ship — deliberately not decided here).

## What was wrong

```
$ bash scripts/npu-driver-reapply.sh          # origin/main, this box
ERROR: …/kernel/drivers/accel/amdxdna/amdxdna.ko.zst not found — is kernel … installed?
exit 1
$ modinfo -n amdxdna
/lib/modules/…/kernel/drivers/accel/amdxdna/amdxdna.ko      # uncompressed
```

The script that exists to put the custom driver back after a kernel upgrade could not run at all: the running kernel ships `amdxdna.ko`, it looked for `amdxdna.ko.zst`. `build/Release` does not exist under `~/xdna-driver` and `/var/lib/dkms/mok.key` is absent while the module loaded here is unsigned, so two more steps would have failed had it got that far.

And the tree it would have rebuilt from is the one upstream deleted — `~/xdna-driver` HEAD `7004f1c` (2026-08-27) predates `813e0bf` (2026-09-01), `src/driver/amdxdna` is tracked and dirty there, and the **installed** module was built from it (compile path in its own bytes, mtime 2026-09-01 11:50) — #2459's finding, with nothing in the script recording any of it.

## What this changes

Discovery and provenance, no new policy:

* **Installed module discovered** via `modinfo -n amdxdna`, then a directory scan, under either compression; the install step writes the form the kernel already uses (a `.ko.zst` written over a `.ko` name would not load).
* **`source_tree_of()`** classifies a checkout `legacy` / `upstream` / **`unknown`** — the last for a non-git tree. The inline test it replaces read that case as "not legacy" and would have built from a tarball extract of the deleted tree.
* **`module_tree_of()`** reads which tree a *built* module came from out of its own bytes (the identification #2459 used on the running driver). The legacy tree is refused, an unreadable compile path is refused with **exit 2**, and `--allow-legacy-tree` is the explicit override.
* `--no-sign` installs unsigned on purpose, verified by comparing the installed bytes with the build, since `modinfo` can only prove a signature; missing MOK files are diagnosed by name instead of failing inside `sign-file`.
* The header now describes this box rather than the layout the script was written for.

## Verification

On this box, after the change (read-only — it refuses before building):

```
installed module: /lib/modules/…/kernel/drivers/accel/amdxdna/amdxdna.ko (legacy tree)
ERROR: /home/bcloud/xdna-driver is the LEGACY out-of-tree driver tree (src/driver/amdxdna is tracked there).
  HEAD: 7004f1c 2026-08-27 shim: allow firmware log and event trace queries on mgmt PF
  Upstream removed that tree in 813e0bf (2026-09-01); …
  Re-run with --allow-legacy-tree only if you have decided to stay on it on purpose.
```

`Testing/npu_driver_reapply_selfcheck.sh` sources the script and exercises the real helpers — 17 cases: module provenance (legacy / upstream / unknown / both→legacy), discovery (either compression, compressed preferred, none → non-zero), the unsigned byte comparison, the checkout classifier, and a check that the helpers are actually **wired** into the script. Red on `origin/main`'s script, green here.

Two things the test found while being written, both fixed: the script had no sourcing guard (so exercising the helpers ran the host work), and its `set -euo pipefail` sat above that guard, which changed the *caller's* shell options and made the selfcheck exit on its first intentionally-failing call.

`bash Testing/run_all.sh` → `27/41 passed, 14 skipped`, **0 failures**, including `✓ npu_driver_reapply`. Wired at the end of the suite, away from the blocks the other open PRs touch.

Not in scope: whether to move this box to the upstream staging driver. That is #2459, and nothing here decides it — the script just stops installing the obsolete tree while calling itself a safety net.

___

### **PR Type**
Bug fix, Tests, Enhancement

___

### **Description**
- Make driver-reapply find the installed module either compression

- Refuse the legacy out-of-tree tree upstream deleted (813e0bf)

- Add provenance helpers and `--allow-legacy-tree` / `--no-sign` overrides

- Add selfcheck exercising helpers; wire it into `Testing/run_all.sh`

___

### Diagram Walkthrough

```mermaid
flowchart TD
  Reapply["scripts/npu-driver-reapply.sh"] -- "modinfo -n amdxdna, then dir scan" --> Discover["find_installed_module: .ko or .ko.zst"]
  Discover -- "module_tree_of reads compile path" --> Gate{"tree legacy or unknown?"}
  Gate -- "yes, without --allow-legacy-tree" --> Refuse["refuse: exit 1 / exit 2"]
  Gate -- "upstream" --> Sign["sign with MOK key, or --no-sign"]
  Sign -- "install in the form the kernel uses" --> Verify["verify signature or byte-compare"]
  Selfcheck["Testing/npu_driver_reapply_selfcheck.sh"] -- "sources helpers" --> Helpers["module_tree_of / find_installed_module / source_tree_of"]
  RunAll["Testing/run_all.sh"] -- "runs" --> Selfcheck
  Selfcheck -- "grep: gate is wired in" --> Reapply
```

<tr>
  <td>
    <details>
      <summary><strong>npu-driver-reapply.sh</strong><dd><code>Discover module, gate on legacy-tree provenance, add sign override</code></dd></summary>
<hr>

scripts/npu-driver-reapply.sh

<ul><li>Replaces the hardcoded <code>amdxdna.ko.zst</code> check with <code>modinfo -n amdxdna</code> <br>plus a <code>.ko.zst</code>/<code>.ko</code> directory scan, and installs in the form the kernel <br>already uses.<br> <li> Adds pure helpers <code>module_tree_of</code>, <code>find_installed_module</code>, <br><code>installed_matches_raw</code> and <code>source_tree_of</code>, gated behind a <code>BASH_SOURCE</code> <br>sourcing guard that defers <code>set -euo pipefail</code> and host work.<br> <li> Refuses a legacy (<code>src/driver/amdxdna</code>) or undeterminable source/built <br>module with exit 1/2 unless <code>--allow-legacy-tree</code>; parses <code>--reload</code>, <br><code>--no-sign</code>, <code>--help</code> and honours <code>NPU_DRIVER_SRC</code>.<br> <li> Moves signing behind an explicit key-presence check and <code>--no-sign</code>, <br>verifying unsigned installs by byte comparison; rewrites the header to <br>this box's discovered state.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2518/files#diff-d07624783f1d417088d2df212092c263d333bfe051f8b54f301472e900ea8f34">+196/-30</a></td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>npu_driver_reapply_selfcheck.sh</strong><dd><code>Selfcheck covers reapply provenance helpers and wiring</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/npu_driver_reapply_selfcheck.sh

<ul><li>New selfcheck that sources the real script and exercises <br><code>module_tree_of</code>, <code>find_installed_module</code>, <code>installed_matches_raw</code> and <br><code>source_tree_of</code>.<br> <li> Uses fixture modules/trees for legacy, upstream, unknown and <br>both-paths cases, plus a host-only real-module classification that <br>skips when absent.<br> <li> Verifies the helpers are actually wired into the script (call sites, <br><code>--allow-legacy-tree</code>, <code>exit 2</code>), failing if a helper is uncalled.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2518/files#diff-a3387524abdc85a9187e271ad45b706b682f7934d7e45d8203dd341a72f076a7">+126/-0</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>run_all.sh</strong><dd><code>Wire reapply provenance selfcheck into run_all</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/run_all.sh

<ul><li>Adds an "npu driver re-apply provenance" block that runs the new <br>selfcheck, prints its output indented, and counts pass/fail.<br> <li> Documents why the case needs no device, build or loaded driver.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2518/files#diff-60779953d5d72f0d8ef3aea52f86b0b365bbdb42457aae2e5955d6320e32eb2c">+16/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Queue prediction from `git merge-tree` (not from reading the diffs) — this entry will be ejected.

**It conflicts with #2510**, which is ahead of it in the queue (position 2), on `Testing/run_all.sh`.
Not because the two changes are related, but because both insert immediately before the same anchor:

```
echo "======================================"
echo "$((total-fail-skip))/$total passed, $skip skipped"
[ "$fail" -eq 0 ] || { echo "$fail FAILURES"; exit 1; }
```

Two insertions in one gap conflict even when the inserted text is unrelated (the merge base has
nothing there, so git sees two independent additions to the same point). Whichever of the two lands
first, the other ejects — and under `grouping_strategy ALLGREEN` an ejected entry costs every PR
grouped with it a rebuild cycle, not just itself.

The remedy is to move the block to a different gap in `run_all.sh` — any stable anchor more than a
few lines from the summary is enough, e.g. just before one of the `# ── <suite> ──` section headers.
Three checks worth making, all cheap and device-free:

```
# 1. it still parses
bash -n Testing/run_all.sh
# 2. the move did not change the file's contents, only their order
#    (compare line multisets of the revision before and after)
# 3. it now merges cleanly against the sibling AND against main
git merge-tree --write-tree origin/main <your-branch>            # expect rc 0
git merge-tree --write-tree <main + #2510> <your-branch>         # expect rc 0
```

For (2): a pure relocation leaves the line multiset identical; if the count of blank lines changes,
that is fine, but anything else changing means the move was not a pure move.

I hit the same class twice today and fixed it the same way — #2507 (a check block inserted in the
same gap as a sibling's) and #2539 (this exact gap, in this exact file, against #2510). Both are
clean now, so after they land this file's summary gap is back to a single writer.

PR #2532, opened 2026-09-18
test(suite): the orphan tripwire passed on the prose that described the orphans

bong-water-water-bong:
### **User description**
**Stacked on #2510** (its branch is locked in the merge queue, so this carries the fix; GitHub retargets it to `main` when #2510 merges).

## Why

The tripwire this PR adds decides whether a `Testing/*_selfcheck.*` file is invoked by grepping the raw text of `run_all.sh` and the workflows for its basename. But the tripwire's own header comment names the two orphans it was written for:

```
# Two selfchecks sat here invoked by nothing (hrx_backend_selfcheck.cpp,
# lse_backend_selfcheck.cpp), and the HRX one had rotted: ...
```

So the sentence describing the defect satisfies the check that looks for it. Measured with a new `Testing/zz_new_gate_selfcheck.py` whose only mention anywhere is a comment:

```
before:  ✓ selfcheck wiring (every Testing/*_selfcheck.* is invoked)      exit 0
after:   ✗ selfcheck wiring: nothing invokes: zz_new_gate_selfcheck.py
```

The realistic version of this is not adversarial: the next person who adds a gate and writes "coming soon: Testing/foo_selfcheck.py" in a comment, or documents an orphan the way this PR documents its own, gets a green tripwire for a file nothing runs.

## What changed

- `strip_comments_stream` removes YAML/shell comments from the corpus before the comparison.
- The comparison is now a function, `selfcheck_orphans <corpus> <file...>`, so the control can run it against a synthesized corpus: the block ends by requiring that a selfcheck named **only** in a comment is reported. If comment-stripping is ever removed, that control fails the run — the blindness cannot return silently.
- The corpus is passed as an argument instead of piped into `grep -q`, which also removes real noise: the old `printf … | grep -q` died of SIGPIPE once per file and printed up to 25 `Broken pipe` write errors on stderr per run.

## Verification

```
as committed:                              ✓ invoked   + ✓ control
two invocations deleted, comments kept:    ✗ reports lse_backend_selfcheck.cpp
                                           (hrx is still named by three other live lines, correctly)
new orphan named only in a comment:        old ✓  →  new ✗ reports it
```

`fail=1` reaches `exit 1` at `run_all.sh:531`, so this is the job's verdict, not just a log line.

## Scope

Whether `run_all.sh` itself is invoked in CI is #2510's business — this only makes the tripwire's answer mean what it says.

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Strip shell/YAML comments before matching selfcheck names

- Extract `selfcheck_orphans` helper; pass corpus as argument

- Add control asserting a comment is not an invocation

- Stop `printf | grep -q` SIGPIPE noise on stderr

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Corpus["run_all.sh + workflows"] -- "strip_comments_stream" --> Stripped["comment-free corpus"]
  Stripped -- "selfcheck_orphans" --> Check{"basename named?"}
  Check -- "yes" --> Pass["✓ selfcheck wiring"]
  Check -- "no" --> Fail["✗ orphan named, fail++"]
  CtlCorpus["synthesized corpus: comment only"] -- "selfcheck_orphans" --> Ctl{"orphan reported?"}
  Ctl -- "no" --> CtlFail["✗ control: check is blind"]
  Ctl -- "yes" --> CtlPass["✓ control: comment is not an invocation"]
```

<tr>
  <td>
    <details>
      <summary><strong>run_all.sh</strong><dd><code>Make selfcheck orphan tripwire ignore comments</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/run_all.sh

<ul><li>Adds <code>strip_comments_stream</code> (sed) to remove shell/YAML comments from <br>the scan corpus, so prose naming an orphan no longer satisfies the <br>tripwire.<br> <li> Refactors the check into <code>selfcheck_orphans <corpus> <file...></code>, taking the corpus as an <br>argument instead of piping into <code>grep -q</code>, which also eliminates <br>per-file SIGPIPE <code>Broken pipe</code> stderr noise.<br> <li> Adds a control block that runs the helper against a synthesized <br>comment-only corpus and fails the run if a comment counts as an <br>invocation.<br> <li> Documents why the raw-text version was blind, citing the measured <br>green result with invocations deleted.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2532/files#diff-60779953d5d72f0d8ef3aea52f86b0b365bbdb42457aae2e5955d6320e32eb2c">+43/-6</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
# Independent reproduction, via a different mutation than the one in your note — and a collateral check

Your note measures the tripwire passing on **prose**: the block's own comment names
`hrx_backend_selfcheck.cpp`, so deleting the invocations left the tripwire satisfied. I hit the same
blind spot from the other direction before finding this PR — **commenting out a live invocation**:

```
$ sed -i 's|^run arch      Testing/arch_mapping_selfcheck.cpp --|# run arch ...|' Testing/run_all.sh
before this PR:  ✓ selfcheck wiring (every Testing/*_selfcheck.* is invoked)      <- blind
with this PR:    ✗ selfcheck wiring: nothing invokes: arch_mapping_selfcheck.cpp  <- caught
```

so the stricter rule covers the disable-by-comment case too, not only the self-describing-comment case.
(`arch_mapping_selfcheck.cpp` is named in exactly one place in the corpus — `run_all.sh:52` — so that
mutation was unambiguous.) Your built-in control at `:60-66` is the part that makes it stay caught, and
it fails in the right direction: it asserts a synthesized comment-only corpus *does* produce an orphan.

## Collateral check on the stricter rule

Stripping comments also drops the documented exemption in the block's own comment — *"either wire it
into this script (or a workflow), or say above why it is manual"*. Under the new rule a note saying
"this one is manual" no longer satisfies the tripwire. Measured across the current tree:

```
35 selfchecks in Testing/
  0 are matched only by a comment        <- nothing becomes an orphan
  0 are matched only via a workflow      <- all 35 are named by run_all.sh itself
```

So the tightening orphans nothing today, and the `or say above why it is manual` sentence is now an
aspiration the check no longer honours. Worth either updating that sentence or adding an explicit
manual list, so the next person who documents a manual selfcheck is not surprised when the tripwire
fails on it — that surprise is the only cost I can see in this change.

bong-water-water-bong:
Merged and verified on the merged tree — the collateral I flagged is nil:

```
$ (the tripwire block, on main 632b82c31)
✓ selfcheck wiring (every Testing/*_selfcheck.* is invoked)
  fail=0
```

That is the stricter rule running against a tree that now also carries the selfchecks added by #2543
(`capture_interposer_build_selfcheck.sh`) and #2544 (`validate_claims_trigger_selfcheck.py`) — both
wired into `run_all.sh` — so nothing became an orphan and the `or say above why it is manual` sentence
is the only loose end, as noted above.

For the record, `main`'s own CI cannot show this: its C++ job has been cancelled in 25 of its last 25
runs (merges arrive faster than a ~13-20 min build), so these were run by hand on a detached worktree.
The other selfchecks touched by the recent merges pass there too — `repo_docs_selfcheck.py`,
`census_scripts_selfcheck.py` (now reporting 6 scripts), `validate_claims_trigger_selfcheck.py`,
`tools/validate_claims.py --check-readme`, and `capture_interposer_build_selfcheck.sh` (its injected
syntax-error control is rejected).

PR #2506, opened 2026-09-17
ci(bench): build the engine through its CMake target, not a copied source list (+ tripwire)

bong-water-water-bong:
### **User description**
**Three commits, one story.** (1) A guard that notices when the engine's source set is a copy that has drifted. (2) The same guard extended to `build_npu.sh`'s `ENGINE_OBJS`, which names three TUs the CMake target never lists. (3) The copy removed — `bench.yml` now builds the engine **through its CMake target**, so a source added to the engine is compiled there by construction. The guard stays as the tripwire for anyone who re-introduces a hand list.

The change to `bench.yml` is verified by CI rather than by argument: this PR's own push ran the benchmark job, the XRT gate opened, `cmake -S engine/npu` configured, the target built, and the binary launched and reached `I8Ctx::init`. It is also cheaper than the hand-rolled `g++` it replaces (17 s against ~30 s) and it stops linking `npu-infer/src/flm_bridge.cpp`, which `CMakeLists.txt` dropped on 2026-08-15. The run also showed the job passing without producing a benchmark — filed as #2508, out of scope here.

---

## Why

`npu_engine_universal.cpp` is compiled **with XRT** by exactly one CI job, and that job lists its sources **by hand**. The required `C++ (cmake configure + build)` job runs without XRT — `ci.yml:38` says *"libxrt-dev not available — fused NPU backends stubbed"* — and `CMakeLists.txt:2351` only adds the engine `if(XRT_FOUND)`:

```cmake
if(XRT_FOUND)
    add_subdirectory(engine/npu EXCLUDE_FROM_ALL)
else()
    message(STATUS "npu: XRT NOT found — skipping NPU engine build (npu_engine_universal will not be a target)")
endif()
```

So on that runner the target **does not exist** and the engine is never compiled: no required check can fail on a missing TU. `bench.yml`'s hand-written `g++` line is the only thing that can, and nothing compared the two descriptions of the build.

They had already drifted. #2435 adds bridge TUs to `NPU_ENGINE_SOURCES` and `build_npu.sh` but never touches `.github/` (`git diff --name-only origin/main...origin/goal/runlist-decode-wire -- .github/` is empty), and the only signal is an optional job's **26 undefined references** link error. Filed as #2505.

## What

[...]

Issue #2484, opened 2026-09-17
bug(npu): 19 of 85 slots have no xclbin — 4 of 18 deployed models fail context init on the legacy path


bong-water-water-bong:
## Summary

A census of the xclbin every deployed model needs (from #2329, tool in #2483) finds **33 of 85 slots have no artifact under any candidate name**, across **7 of the 18 models** in `~/.config/flm/models`. On the default (legacy) NPU path a missing xclbin is fatal, not degraded — so pointing `npu_engine_universal` at any of those seven fails at context init.

```
$ python3 Testing/npu_xclbin_census.py
18 models, 85 slots checked (4 where GU is fused, 5 where it is split)
  — 33 slot(s) with no xclbin under any candidate name
```

| model | missing slots |
|---|---|
| Gemma3-1B | 4 (QKV, O, GU, D) |
| Gemma3-4B | 5 (QKV, O, G, U, D) |
| LFM2-1.2B | 5 |
| LFM2-2.6B | 5 |
| Llama-3.1-8B | 5 |
| Llama-3.2-1B | 5 |
| Qwen3-1.7B | 4 |

Gemma3, LFM2 and Qwen3-1.7B have **no artifact of any name**; the misses are `final_i8_QKV_K1152_N1536.xclbin`, `final_i8_GU_K1152_N13824.xclbin`, `final_i8_QKV_K2048_N4096.xclbin` and so on — shapes nothing else in the set shares. The other 11 models resolve fully, via the model tag, the dimension-keyed name, or (Qwen3-4B) both in one context.

## Why this is a failure and not just a missing file

Two independent checks, because "missing artifact" could plausibly mean "silently slower":

1. **The shapes are right.** The census derives them from the engine's own formulas (`npu-infer/include/model_config.h:262-279`) and reproduces the one case the engine records — Qwen3-4B → `final_i8_G_K2560_N9728.xclbin`, the artifact #2329 was filed about.
2. **A missing xclbin is fatal on this path.** `engine/npu/src/npu_engine_universal.cpp:28` includes `npu_engine_i8ctx_inc.h`, and **both** of that `I8Ctx`'s xclbin load sites return false:

   ```
   107 / 194:  xc = std::make_unique<xrt::xclbin>(std::string(xp));
   112 / 199:  fprintf(stderr, "  I8Ctx: xclbin/kernel init failed: %s\n", ex.what());
   113 / 200:  return false;
   ```

[...]

PR #2538, opened 2026-09-18
ci(bench): the NPU Benchmark job could conclude success without a benchmark

### LINKED ISSUE #2508: ci(bench): the benchmark job concludes success without producing a benchmark — 'No ms/tok line found' is a warning, not a failure
## Summary

The `NPU Benchmark` job can conclude **success without producing a benchmark**. A run just did exactly that:

```
I8Ctx: xclbin/kernel init failed: No such file
  '/home/bcloud/.local/share/1bit-monster/xclbins/final_i8_QKV_K1024_N4096.xclbin'
FAIL QKV
---
##[warning]No ms/tok line found
```

The job exits 0. The only signal is a warning annotation, and `##[warning]` does not fail a step.

## Why it is worth an issue rather than a note

The step's tail is:

```bash
OMP_NUM_THREADS=16 ./npu_engine_universal "$MODEL_PATH" "$DEC" 2>&1 | tee benchmark.log
grep "ms/tok" benchmark.log || echo "::warning::No ms/tok line found"
```

So the last command is a `grep` whose failure is swallowed by `|| echo`, and the step's exit status is that of `echo`. Every way the benchmark can fail to happen — no model, no xclbins, a driver hiccup, a crash before the timing line — lands in the same place: a green job.

This is the repo's own stated pattern, from #2424: *"a gate that never ran is not a pass."* The job is named for a measurement, and right now its green says "the benchmark ran" when it may only say "the workflow was syntactically fine".

## What it costs

For a run where the engine cannot start, the benchmark job's green is indistinguishable from a run that measured 40 tok/s. Anything reading job status — a release gate, a badge, a human scanning Actions — reads a benchmark that did not happen as a benchmark that passed.

## Options

1. **Fail when no measurement was produced.** `grep -q "ms/tok" benchmark.log || { echo "::error::the benchmark produced no ms/tok line"; exit 1; }`. Simplest, and matches the "fail closed" line the rest of this repo takes.
2. **Fail only when the engine did not start**, and keep a skip-with-warning for a missing model — i.e. distinguish "no model available here" (an environment fact) from "the model was there and the engine failed". That needs the two cases to be told apart explicitly rather than both collapsing into a missing grep.
3. **Annotate the summary with which case it was**, and keep exit 0 — the weakest option, but better than a bare warning.

1 is what this issue recommends unless the maintainer wants the job to tolerate a bare runner.

## Context

Found while verifying a change to this same job's build step (#2506): the run proved the new CMake build works — the binary built, launched, and reached `I8Ctx::init` — and then proved this, which is not what the run was looking for.

bong-water-water-bong:
### **User description**
Fixes #2508.

## Why

The `NPU Benchmark` job could conclude **success without producing a benchmark**. Four ways, three of them silent:

```yaml
      - name: Check NPU driver
        run: |
          if ! ls /dev/accel/accel* >/dev/null 2>&1; then
            echo "::warning::NPU device not found — driver may need reload"     # job stays green
            echo "DRIVER_OK=false" >> $GITHUB_ENV
      ...
      - name: Benchmark
        if: env.DRIVER_OK == 'true'                                            # skipped, job still green
        run: |
          OMP_NUM_THREADS=16 ./npu_engine_v12 "$MODEL_PATH" "$DEC" 2>&1 | tee benchmark.log
          grep "ms/tok" benchmark.log || echo "::warning::No ms/tok line found" # step status = echo's
```

1. **`| tee` discards the engine's status** — a pipeline reports the last command's, which is `tee`'s. The engine does report this failure: every `init_i8`/`init_bf16` failure is `return 1`, and the success path ends in `_exit(0)`. The log quoted in the issue (`I8Ctx: xclbin/kernel init failed: No such file`, then `FAIL QKV`) exited **1** into a green job.
2. **`grep || echo` discards the measurement's status**, so "no measurement was produced" was an annotation, not a verdict.
3. **The device and XRT preconditions only warn**, so the step that does the work is skipped and the job still goes green: a job named `NPU Benchmark` reporting success with nothing benchmarked — indistinguishable from a measured 40 tok/s.
4. The model check already failed correctly and still does.

## What changed

* both preconditions exit 1 with an `::error::` naming the reason (device absent, XRT absent);
* the engine's status is captured through `PIPESTATUS` and fails the step when non-zero;
* the measurement is required: no `ms/tok` line is an error, which is the line this repo already states for itself in #2424 — *"a gate that never ran is not a pass"*;
* the step ends by printing the measured lines, so the number is at the bottom of the log instead of somewhere above.

## Verification

A stub engine, extracted step scripts, and **no NPU use** — this box is single-device and other lanes are on it (coordination rule 4), so nothing here touches the device:

```
STUB_MODE=ok        exit 0   prints "=== 42.1 ms/tok (23 tok/s) | boot=900ms batches=4 tokens=32 ==="
STUB_MODE=initfail  exit 1   "the engine exited 1 without completing a benchmark — the log above says why"
STUB_MODE=noline    exit 1   "the engine produced no ms/tok line — no benchmark was measured"
device present      exit 0   DRIVER_OK=true
device absent       exit 1   ::error::NPU device not found — no benchmark can be measured on this runner (#2508)
XRT present         exit 0   XRT_OK=true
XRT absent          exit 1   ::error::XRT not found in ldconfig — the engine cannot be built or run here
```

Each was run from the script as extracted from the YAML (not a paraphrase), with a stubbed `ls`/`ldconfig` for the absent cases.

`git merge-tree --write-tree` against the queued #2506 — which edits this same file's build step — is **clean**, so the two can land in either order.

## Note on tolerance

Failing when the runner has no device/XRT is the fail-closed reading of *"a gate that never ran is not a pass"*. If a bare runner should instead be tolerated, that is one `continue-on-error` on those two checks (or a `workflow_dispatch` input) — but it should be an explicit choice with a named reason, not the current silence.

___

### **PR Type**
Bug fix, Tests, Enhancement

___

### **Description**
- Pin the NPU benchmark job to `[self-hosted, strix-halo]`

- Fail the job when the NPU driver or XRT is absent

- Require an `ms/tok` measurement instead of warning

- Retry candidate models with `NPU_XCLBIN_DIR` set

- Add `runner_label_selfcheck.py` with fixture controls

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["bench.yml: runs-on self-hosted"] -- "matched XRT-less runner, steps skipped" --> B["job reported SUCCESS with no benchmark (#2508)"]
  A -- "pin" --> C["runs-on [self-hosted, strix-halo]"]
  C --> D["Check NPU driver / Check XRT"]
  D -- "device or XRT missing" --> E["::error:: + exit 1"]
  D -- "present" --> F["Benchmark: candidate model loop"]
  F -- "PIPESTATUS != 0 or no ms/tok" --> G["::error:: + exit 1"]
  F -- "xclbin No such file" --> H["warning, try next candidate"]
  H --> F
  F -- "ms/tok found" --> I["print measurement lines"]
  J["Testing/runner_label_selfcheck.py"] -- "fixtures: violating vs correct" --> K["flag bare label with hardware marker"]
  J -- "scan real .github/workflows" --> L["Testing/run_all.sh runner_label gate"]
```

<tr>
  <td>
    <details>
      <summary><strong>runner_label_selfcheck.py</strong><dd><code>Add runner label self-check with fixture controls</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/runner_label_selfcheck.py

<ul><li>Adds a new self-check that flags any workflow doing NPU/XRT work whose <br>runner is selected by the bare <code>runs-on: self-hosted</code> label<br> <li> Detects hardware work through <code>HARDWARE_MARKERS</code> (<code>xrt_coreutil</code>, <br><code>/dev/accel</code>, <code>npu_engine_universal</code>, <code>xclbin</code>, <code>strix-halo</code>) and matches <code>BARE</code> <br>vs <code>PINNED</code> <code>runs-on</code> regexes<br> <li> Builds its own fixtures (<code>violating</code>, <code>correct</code>, <code>correct_unrelated</code>) and <br>requires three controls to hold before scanning the real tree<br> <li> Scans <code>.github/workflows/*.yml</code>/<code>*.yaml</code>, reports each violation with the <br>fix (<code>runs-on: [self-hosted, strix-halo]</code>) and exits 1 on any failure</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2538/files#diff-c0e1c176a2d3086b0d9da8fe3a48130ec937d582836972dd93aaeb61f2313348">+143/-0</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>run_all.sh</strong><dd><code>Wire runner label self-check into run_all.sh</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/run_all.sh

<ul><li>Wires the new <code>Testing/runner_label_selfcheck.py</code> into the suite as a <br><code>runner_label</code> gate<br> <li> Prints check output indented on success; on failure tails the last 12 <br>lines and increments <code>fail</code><br> <li> Documents that bare <code>self-hosted</code> let the benchmark job skip its work <br>and report success</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2538/files#diff-60779953d5d72f0d8ef3aea52f86b0b365bbdb42457aae2e5955d6320e32eb2c">+16/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>bench.yml</strong><dd><code>Make benchmark job fail without a measurement</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/bench.yml

<ul><li>Pins the job to <code>runs-on: [self-hosted, strix-halo]</code> so it cannot be <br>scheduled onto the runner without XRT, with the measured false-green <br>runs cited in the comment<br> <li> Turns the "Check NPU driver" and "Check XRT" warnings into <code>::error::</code> <br>messages followed by <code>exit 1</code><br> <li> Sets <code>NPU_XCLBIN_DIR</code> to the repo's <code>engine/npu/xclbins</code> tree and iterates <br>over all present candidate models instead of picking the first <br>existing file<br> <li> Captures the engine status via <code>PIPESTATUS</code>, fails when no <code>ms/tok</code> line <br>is produced, skips only candidates missing an xclbin, and prints the <br>measured lines at the end</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2538/files#diff-a5e740be96415373789689f814583e93ff2a8f05eae6481e94505fd6cb6bc6a7">+83/-13</a>&nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
# The exact mechanism, from the resolver — it also explains why the repo-relative fallback missed

I had written "the engine resolves through an installed layout, which is empty". Reading
`engine/npu/src/npu_paths.h:83` makes it precise, and it confirms the fix rather than assuming it:

```cpp
inline std::string npu_xclbin_dir() {
    const char* e = getenv("NPU_XCLBIN_DIR");
    if (e && e[0]) {
        if (npu_dir_exists(e)) return e;                 // an override that does not exist is not an override
        npu_warn_ignored("NPU_XCLBIN_DIR", e, "engine/npu/xclbins");
    }
    if (npu_dir_exists("engine/npu/xclbins")) return "engine/npu/xclbins";   // <-- relative to the CWD
    const char* home = getenv("HOME");
    if (home && home[0]) return std::string(home) + "/.local/share/1bit-monster/xclbins";
    return "/usr/local/share/1bit-monster/xclbins";
}
```

Three consequences for this job:

1. **The repo-relative fallback is CWD-relative**, and the benchmark step does `cd engine/npu/build-cmake`
   before running the binary. From there `engine/npu/xclbins` does not exist, so the fallback that would
   have found the repo's 374 artifacts is skipped — which is precisely how the job ended up in the
   installed layout.
2. **That layout exists but is empty** (`ls ~/.local/share/1bit-monster/xclbins | wc -l` = 0), so the
   resolver returns a real directory containing nothing, and every candidate model dies at `FAIL QKV`.
   An empty directory is a worse outcome than a missing one here: nothing in the log says the path was
   wrong, only that the artifacts are absent.
3. **The override is honoured only when the directory exists** — and after checkout it does, at
   `${{ github.workspace }}/engine/npu/xclbins`. If that path were wrong the log would say so explicitly
   (`npu_warn_ignored` names the ignored value and the fallback), which is why setting the variable is a
   safe way to fix this: a mistake in it is loud rather than silent.

Worth noting as an alternative I did not take: dropping the `cd` and invoking
`engine/npu/build-cmake/npu_engine_universal` from the repo root would let the relative fallback hit, with
no environment change at all. The explicit variable is less clever and survives someone later adding
another `cd` — but it is a choice, not the only correct one.

bong-water-water-bong:
# Executed, not read: the resolver reproduces the defect and the fix, on this box

Following the comment above, I compiled a probe against the real header (`npu_paths.h`, which is
`#pragma once` + inline and includes standalone) and ran `npu_xclbin_dir()` in the three conditions this
job moves between. No device involved:

```
1. from the repo root
   cwd has engine/npu/xclbins: yes
   npu_xclbin_dir() -> engine/npu/xclbins

2. from a cwd WITHOUT the repo layout      (what `cd engine/npu/build-cmake` produces)
   cwd has engine/npu/xclbins: no
   npu_xclbin_dir() -> /home/bcloud/.local/share/1bit-monster/xclbins      <- 0 files on disk

3. same cwd, with the variable this PR sets
   npu_xclbin_dir() -> /home/bcloud/1bit-MONSTER/engine/npu/xclbins        <- 129 xclbins

4. control: NPU_XCLBIN_DIR=/nope/not/here
   NPU: NPU_XCLBIN_DIR='/nope/not/here' is not a directory on this machine — ignoring it and
        using 'engine/npu/xclbins' instead (unset or fix the variable)
   npu_xclbin_dir() -> /home/bcloud/.local/share/1bit-monster/xclbins
```

Case 2 is the failure in run 35299693789 reproduced from the code path rather than argued: an empty
directory is returned, so every `init_i8` fails on a file that is absent from it and the log says
`FAIL QKV` without ever naming a wrong path. Case 3 is this PR's one-line change doing what it claims.
Case 4 is why I am comfortable with it: the variable is not trusted blindly — a wrong value is ignored
**and named**, so a typo here would appear in the log instead of silently reverting to the empty layout.

What this does not and cannot show is whether the engine then *measures* on the NPU — that needs the
runner, and the queued `NPU Benchmark` run for this commit is the end-to-end answer. Everything up to
that boundary is now demonstrated rather than inferred.

bong-water-water-bong:
# End-to-end: the job measures — `439.7 ms/tok (2 tok/s)` — and the cause was the search path, not missing artifacts

Run `35301292724` (`e38456c12`) completed **success**, on `strixhalo-pr-agent`:

```
5. Build engine (continuous batch): success
6. Benchmark: success
...
I8Ctx::init xp=<repo>/engine/npu/xclbins/final_i8_QKV_qwen3_0_6b.xclbin
=== 439.7 ms/tok (2 tok/s) | boot=7ms batches=31 tokens=32 ===
Model: /home/bcloud/.config/flm/models/Qwen3-0.6B-NPU2/model.q4nx
```

It measured the **first** candidate, having previously died on all four at `FAIL QKV`. This is the
end-to-end answer to the one thing the probe could not show: the engine now initialises and produces a
number where it used to produce nothing but a green check (#2508).

**And it corrects a claim of mine that was in this PR and in two issues.** The engine resolved a
**tag-keyed** artifact — `final_i8_QKV_qwen3_0_6b.xclbin` — not a dimension-keyed one. All four of that
model's artifacts are in the repo under tag names:

```
final_i8_QKV_qwen3_0_6b.xclbin    tracked
final_i8_O_qwen3_0_6b.xclbin      tracked
final_i8_GU_qwen3_0_6b.xclbin     tracked
final_i8_D_qwen3_0_6b.xclbin      tracked
dimension-keyed `K1024` names:    0 tracked
```

What does not exist is the *dimension-keyed spelling* `final_i8_QKV_K1024_N4096.xclbin` — which is the
name I grepped for when I wrote that this model "has no committed artifact" and added `K=1024` to
#2484's missing-shape inventory. Both were wrong, and they pointed at a build when the problem was a
path. The comment in `bench.yml` is corrected in the commit after this one.

So the accurate story is the one the probe demonstrated: `cd engine/npu/build-cmake` defeats the
CWD-relative fallback, the resolver falls through to an installed layout that exists and is empty, and
every candidate fails on files that are actually present in the checkout. `NPU_XCLBIN_DIR` fixes it in
one line, and a wrong value would be named in the log rather than silently falling back.

bong-water-water-bong:
# The silent-skip class is confined to this file, so the fix is not partial

Worth stating because it is the kind of thing that turns out to be true in three other workflows. A
search for steps guarded by an environment flag:

```
$ grep -rn "if: env\." .github/workflows/*.yml
  bench.yml:57:  if: env.XRT_OK == 'true'
  bench.yml:63:  if: env.XRT_OK == 'true'
  bench.yml:78:  if: env.DRIVER_OK == 'true'
  bench.yml:123: if: env.DRIVER_OK == 'true'
```

Four guards, all in `bench.yml`, over exactly two flags — and on `main` both flags are still set behind a
warning while the job reports success:

```
bench.yml:39   echo "::warning::NPU device not found — driver may need reload"
bench.yml:49   echo "::warning::XRT not found in ldconfig — NPU benchmarks unavailable on this runner"
```

This PR is what turns both into `::error::` + `exit 1`, and the runner pin is what makes that legitimate
rather than a misfire: on `[self-hosted, strix-halo]` a missing driver or a missing XRT is a real fault,
whereas on the bare `self-hosted` label it was the wrong box being selected (and skipping its own work
quietly, 42 of 60 runs). So there is no second instance of this pattern elsewhere in the workflow set to
fix in a follow-up — the class is these four steps, and they are the ones this PR changes.

bong-water-water-bong:
Evidence boundary, stated precisely so nobody has to guess which commit was measured:

- The end-to-end verification (run 35301292724, `conclusion=success`, log line
  `=== 439.7 ms/tok (2 tok/s) | boot=7ms batches=31 tokens=32 ===` on Qwen3-0.6B-NPU2,
  runner `strixhalo-pr-agent`) is for head **e38456c12**.
- The current head is **9eef07b0f**, whose only delta is comment-only:

      $ git diff --stat e38456c12 9eef07b0f
       .github/workflows/bench.yml | 22 ++++++++++++++++------
       1 file changed, 16 insertions(+), 6 deletions(-)
      $ git diff -U0 e38456c12 9eef07b0f | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*#'
      (no output)

  Every added/removed line in that delta is a YAML comment, so the measured
  behaviour carries over — but the *check run* on the head is not itself a
  measurement.

- On the head, `benchmark` is `conclusion=cancelled` (job 105468307058,
  03:17:27 -> 03:19:28, `runner_name` empty), not `failed`. `benchmark` is not one
  of the four required checks (`C++ (cmake configure + build)`, `Lint
  (clang-format)`, `Scope Guard`, `Version consistency`), so this does not gate the
  merge queue and I am not re-running it just to turn a grey check green.

Related, measured in the same sweep and left as evidence on #2552: on `main`,
`bench.yml` is the only NPU-referencing workflow that does not name the box
(`runs-on: self-hosted` vs `[self-hosted, strix-halo]` in the other four), and bare
`self-hosted` also matches the ryzen runner, which has no NPU. That is the half of
the problem this PR fixes.

PR #2541, opened 2026-09-18
ci: ccache for the C++ build — 13.1 of the required job's 15.8 minutes is compilation

bong-water-water-bong:
## Why

The required `C++ (cmake configure + build)` job is the only long pole in CI, and the merge queue re-runs it for every speculative batch. Measured on `merge_group` run **35292288821**, 15.8 min in total:

| step | time |
| --- | --- |
| Run actions/checkout | 1.1 min |
| Install build deps | 0.4 min |
| Cache TheRock ROCm SDK | 0.5 min — **a hit** (the install step is 0.0 min) |
| Build llama.cpp ggml-vulkan libs | **3.9 min** |
| CMake configure + build (20 targets) | **9.2 min** |
| registry checks + ctest + ShellCheck + CLI smoke | < 0.6 min combined |

So 13.1 of 15.8 minutes is compiling sources that were compiled identically a few minutes earlier in another speculative branch. The TheRock entry already proves the mechanism works on those refs — a cache written on the default branch is visible to `gh-readonly-queue/*` — which is why that restore is a hit.

## What changed

* `ccache` added to the build deps;
* `actions/cache` for `~/.cache/ccache`, keyed per run with a `restore-keys: ccache-${{ runner.os }}-` prefix so the newest cache wins (same shape as the TheRock key: no branch component, so default-branch caches serve the queue's ephemeral refs);
* `-DCMAKE_C_COMPILER_LAUNCHER=ccache`, `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache` and `-DCMAKE_HIP_COMPILER_LAUNCHER=ccache` on **both** configure invocations — llama.cpp's and the main build's;
* `ccache --max-size=1G` and `CCACHE_COMPRESS=1` (persisted through `$GITHUB_ENV`, so it reaches the compiler steps), keeping this inside the repo's 10 GB cache budget beside a ~6 GB TheRock entry;
* `ccache --show-stats` after each build, so the hit rate is in the log rather than assumed.

## What this does not claim

**The size of the speedup is not measured yet.** The local cold/warm A/B is running, but this box is at load 28 from the other lanes, so its absolute timings would not be CI-representative. What is measured, on the runner, is the breakdown above: the job is compilation-bound and everything else is under a minute.

The verification that matters is the hit rate in this PR's own C++ run — and in the runs after it. If it stays near zero, the change is a cache step that costs 0.5 min and buys nothing, and it should be reverted; the stats line is there so that is a fact rather than an argument.

## Risk

Low and reversible: ccache is content-addressed (a wrong hit is a compiler-hash collision, not a warm-cache hazard), it only adds a launcher to the existing commands, and reverting is deleting the cache step and the three `-D…_LAUNCHER` flags. The one real interaction is the shared 10 GB cache budget, which the 1 GB cap is there to respect.

___

### **PR Type**
Enhancement

___

### **Description**
- Add `ccache` to the C++ job's build dependencies

- Cache `~/.cache/ccache` with a per-run key and restore-keys

- Wire ccache launchers into llama.cpp and main CMake configures

- Print `ccache --show-stats` after each build

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Deps["Install build deps (+ccache)"] -- "installs" --> Ccache["Ensure/Configure ccache step"]
  Ccache -- "cache key restore-keys" --> Store["actions/cache ~/.cache/ccache"]
  Ccache -- "sets CCACHE_COMPRESS + 1G" --> Configure["cmake configure (llama.cpp + main)"]
  Configure -- "-DCMAKE_*_COMPILER_LAUNCHER=ccache" --> Build["Ninja build steps"]
  Build -- "ccache --show-stats" --> Log["Hit rate in CI log"]
```

<tr>
  <td>
    <details>
      <summary><strong>ci.yml</strong><dd><code>Add ccache caching and compiler launchers to CI</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/ci.yml

<ul><li>Add <code>ccache</code> to the <code>apt-get install</code> list of the C++ job's build deps.<br> <li> Add an <code>actions/cache</code> step for <code>~/.cache/ccache</code> keyed per <code>github.run_id</code> <br>with <code>restore-keys: ccache-${{ runner.os }}-</code>, plus a configure step <br>running <code>ccache --max-size=1G</code>, <code>--zero-stats</code> and exporting <br><code>CCACHE_COMPRESS=1</code> to <code>$GITHUB_ENV</code>.<br> <li> Pass <code>-DCMAKE_C_COMPILER_LAUNCHER=ccache</code> / <code>CXX</code> to the llama.cpp <br>configure and add <code>CXX</code>/<code>HIP</code> launchers to the main <code>cmake</code> configure, with <br><code>ccache --show-stats</code> printed after each build on both paths.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2541/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f">+36/-3</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Verification status, so the "in flight" line in the description is accurate rather than stale:

**Wiring — verified.** A local configure with the launcher flags produces a ninja file that routes compiles through ccache: **2,710** `ccache` occurrences in `build.ninja`, and `CMakeCache.txt` records `CXX_COMPILER_LAUNCHER:UNINITIALIZED=ccache`. (cmake 4.2.3 here, 3.28 on `ubuntu-24.04`; both support the launcher variables.)

**The cold/warm A/B — abandoned deliberately.** This box is at load 28 from the other lanes, so its absolute build times are not CI-representative, and running a full C++ build on it to produce a number I would have to caveat twice is worse than not running it. I killed it and cleaned the scratch worktree and ccache directories.

**What replaces it.** This PR's own C++ run has the `ccache` and `Configure ccache` steps in its plan (they are scheduled in run `35294215297`), and the build prints `ccache --show-stats`. So the sequence is: this run shows the cold rate (near-zero hits, `ccache --version` proving the step ran), and the runs after it show whether the cache is being used. If the hit rate stays near zero, the honest response is to remove this — the step costs ~0.5 min and buys nothing, and the stats line exists so that is a fact rather than an argument.

bong-water-water-bong:
First numbers from this PR's own run (`35294215297`, C++ job in progress), so the "in flight" note above has data behind it:

| step | this run (cold ccache) | the citable run `35292288821` |
| --- | ---: | ---: |
| `ccache` (restore) | 0.0 min | — |
| `Configure ccache` | 0.0 min | — |
| Build llama.cpp ggml-vulkan libs | **5.5 min** | 3.9 min |
| CMake configure + build | in progress | 9.2 min |

So the cold run is **slower** on the first build step (5.5 vs 3.9), which is the expected shape: every translation unit is hashed and stored with nothing to hit, and the cache is empty on a brand-new key. The `ccache` and `Configure ccache` steps themselves are 0.0 min.

That makes the next runs the ones that decide this: with the cache populated, `llama.cpp` and the 20-target build should drop, and if the hit rate stays near zero the honest response is still to remove the step. The `ccache --show-stats` line after each build is where that shows up.

bong-water-water-bong:
Measured, and the verdict is keep. The same run, second attempt, same `run_id` and therefore the same cache key:

| step | cold (attempt 1) | warm (attempt 2) |
| --- | ---: | ---: |
| Build llama.cpp ggml-vulkan libs | 5.5 min | **1.6 min** |
| CMake configure + build (20 targets) | 13.7 min | **1.1 min** |
| **C++ job total** | **21.3 min** | **5.2 min** |

```
attempt 1  Cacheable calls 377  Hits   0 / 377 (  0.00%)   # llama.cpp step
           Cacheable calls 753  Hits   0 / 753 (  0.00%)   # cmake step
attempt 2  Cacheable calls 377  Hits 377 / 377 (100.0%)
           Cacheable calls 753  Hits 753 / 753 (100.0%)
Local storage: Cache size 0.046 GB / 1.0 (4.62%)
```

Three things this settles:

1. **ccache matches this toolchain, HIP included.** 753/753 is the whole CMake build, which is the HIP-heavy half (94 `.hip` sources); nothing falls back to a miss, so the one real risk of the idea — HIP-through-`amdclang++` not caching — is falsified.
2. **The warm job is faster than the fastest baseline I sampled.** Baseline C++ totals over recent main runs were 9.3, 14.6, 3.2, 2.3 and 0.7 min (the small ones skipped the XRT branch), with cmake 2.5-7.8 min. This run's 5.2 min total, cmake 1.1, sits below all of them.
3. **The cold penalty amortises on the next run.** It is real — attempt 1 was 21.3 min against that 9.3-14.6 range, and I could not separate it from a slow instance because a rerun lands on a different one — but a queue batch of five runs now costs roughly `21.3 + 4×5.2 ≈ 42 min` against `5×12 ≈ 60`, before counting the fact that most of those runs are the same sources.

The cache stays small (0.046 GB, 4.62% of the 1 GB cap), so saving and restoring it is cheap in the budget sense too.

I am keeping the change on that evidence. The honest caveat is the cold number's attribution: with one cold sample on a different instance I cannot prove ccache caused the +7 min rather than the instance being slow, and if a future cold run is again far above the baseline range, the first thing to drop is the HIP launcher — the C/C++ launchers would keep the llama.cpp half.

bong-water-water-bong:
# Where this change's effect is actually visible — and where it can never be

Follow-up measurement while waiting for a post-merge build to finish. Two things I had wrong about
the baseline, both worth recording so the next person measures the right thing.

**1. `main`'s own C++ job has not completed once in the last 25 runs.** All 25 are `cancelled`:

```
02:54:59 AFTER  in_progress   run 35301207327
02:47:40 AFTER  cancelled  7.2 min
02:44:24 AFTER  cancelled  3.3 min
02:43:16 before cancelled  0.8 min
02:25:40 before cancelled 12.7 min
...
```

`ci.yml` cancels in progress, and main receives merges faster than a ~13-minute build finishes, so the
job is killed every time. That is not a defect — the merge queue already builds the speculative
post-merge state on `gh-readonly-queue/main/pr-N-…` refs, and those *do* complete — but it does mean
**this change's effect will never show up on main's runs**, which is where I said to look.

**2. The builds that matter take ~20 minutes, not ~13.** The merge-queue C++ jobs, which are the ones
the queue actually waits on, from before this landed:

```
02:12:09  12.9 min      02:25:42  21.1 min
02:17:49  20.1 min      02:38:31  16.2 min
02:20:16  20.4 min      02:41:04  15.7 min
02:22:44  19.9 min
02:23:22  20.8 min        n=8, median ~20 min
```

So the number to compare against is **~20 min per merge-queue build**, and every minute off it is a
minute off the merge queue's batch time — not just off a PR's checks. Three merge-queue runs since
02:44:25 are still in progress; I will report the comparison when one completes, and if it is not
clearly below the range above, the HIP launcher is the first thing to drop, as noted in the PR body.

bong-water-water-bong:
# Interim: the first cross-run measurement shows **no improvement** — and the stats line should say why

My own PR's premise was cross-run hits. Measuring the same step before and after, on merge-queue builds:

```
                              before ccache    after ccache
Build llama.cpp ggml-vulkan     4.82 min         5.08 min        <- no improvement
CMake configure + build         5.47 min         (still running)
job total                       ~20 min          >= 21 min
```

The earlier evidence for this change was a **within-run** warm rerun (100% hits on the same `run_id`,
which is the same cache key). That is a real measurement of ccache working, but it is not a measurement
of the thing the PR claims, which is that a *later* run is faster. On the first cross-run comparison it
is not. I would rather say that now than let the earlier number stand as if it covered this case.

Two candidate causes, and the `ccache --show-stats` line this PR added after each build will distinguish
them — it is only readable once the job completes, which is why this is an interim note:

1. **The 1 GB cap is too small for this build, so entries are evicted before the next run asks for
   them.** I chose `--max-size=1G` to stay inside the repo's cache budget beside TheRock, and the
   llama.cpp + engine object set is plausibly larger than that. If this is it, the stats will show a
   large `cache size` at the cap with `miss` ≈ `cacheable` and few `hit`s, and the fix is a bigger cap
   or dropping the HIP launcher first (what the PR body already commits to trying first if the numbers
   do not justify the change).
2. **The restore is not reaching the previous run.** The key is
   `ccache-${{ runner.os }}-${{ github.run_id }}` with `restore-keys: ccache-${{ runner.os }}-`, the
   rolling-cache idiom — a per-run key always key-misses and relies on the prefix restore. If that
   restore is failing (path, permissions, or the save never happening), the stats would show
   `cacheable` ≈ `miss` with the cache directory *cold at start*, which is distinguishable from (1) by
   the `cache size` figure: at the cap means evictions, near zero means nothing was restored.

Either way the answer is in a log line, and I will report it rather than infer it. If it comes back as
(1) with the cap at fault, the honest resolution is to shrink this PR's claim, not to keep the cap and
call the change a win.

bong-water-water-bong:
# Root cause: there was nothing to restore, because the cache is only saved by a job that *completes*

The stats from the first completed post-merge build (run 35300535012, 21.4 min) answer it, and they
refute the cause I suspected:

```
Cacheable calls:   753 / 753 (100.0%)
  Hits:              0 / 753 ( 0.00%)
  Misses:          753 / 753 (100.0%)
Local storage:
  Cache size (GB): 0.0 / 1.0 ( 2.11%)      <- after the llama.cpp build: ~21 MB
  Cache size (GB): 0.0 / 1.0 ( 4.62%)      <- after the engine build:    ~46 MB
```

**Not the 1 GB cap.** Both builds together filled 4.6% of it; nothing was evicted. My `--max-size=1G`
was the wrong suspect and I said so before I had the data, which is the only reason it is worth
correcting explicitly now.

**What it is**, from the same log, three lines that belong together:

```
02:46:14  path: ~/.cache/ccache
02:46:14  key: ccache-Linux-35300535012
          (no "Cache restored from key:" line for ccache — contrast TheRock, which has one:
           02:46:14  Cache restored from key: rocm-therock-Linux-2026-W38)
03:05:47  Cache saved with key: ccache-Linux-35300535012
```

The restore ran and found nothing; the save happened only at the very end of this job, in its post-step.
So the cache was empty because **no previous run had ever saved one** — and the reason is in the numbers
from this morning: `main`'s C++ job was cancelled 25 times out of 25, and a cancelled job does not run
its post-step. The cache is populated only by builds that finish, and almost none finish. That is why
this change has produced 0 hits in every run so far: there has never been a cache to hit.

**A prediction, so the explanation is testable rather than plausible:** this run saved
`ccache-Linux-35300535012` at 03:05:47. Any build that *starts after that time* should show hits > 0;
any build that was already in flight before it (the three still running, started 02:47–02:57) should
show 0, because their restore happened before the save existed. If a later build still shows 0 hits with
a non-empty cache, the explanation is wrong and the restore path is at fault instead.

**Where that leaves the change:** the mechanism works (a within-run rerun is 100% hits) but the *key
strategy* cannot deliver cross-run reuse under this workload — a per-run key plus a prefix restore needs
some run to have completed and saved, and on this repo most do not. A key that is stable across runs
(e.g. week-scoped, like the TheRock cache a few lines above: `rocm-therock-Linux-2026-W38` — which
restores reliably every time) is what makes the save reusable the moment one build finishes. Until that
is changed I would not claim this PR saves anything, and if the next completed build contradicts the
prediction above, the honest fix is to revert rather than to keep tuning it.

bong-water-water-bong:
# The prediction failed — reverting, as I said I would

The discriminating test came back negative, and it fails on every ref rather than in one place:

```
key: ccache-Linux-35301936250            <- restore attempted at 03:06:27
(no "Cache restored from key: ccache-…")  <- and found nothing
Cache saved with key: ccache-Linux-35301936250   <- only at the end, 03:21:03

8 completed C++ jobs after the change, merge-queue refs AND PR branches:
  03:09:07 gh-readonly-queue/main/pr-2531-e8f   restored=no   0 / 377 hits
  03:06:24 gh-readonly-queue/main/pr-2549-abd   restored=no   0 / 377 hits
  02:57:21 gh-readonly-queue/main/pr-2518-314   restored=no   0 / 377 hits
  02:56:38 goal/runlist-decode-wire             restored=no   0 / 377 hits
  02:56:29 ci/auto-enqueue-stacked-prs          restored=no   0 / 377 hits
  02:56:20 ci/required-check-names              restored=no   0 / 377 hits
  02:55:12 fix/bench-path-coverage              restored=no   0 / 377 hits
  02:55:01 gh-readonly-queue/main/pr-2547-632   restored=no   0 / 377 hits
```

So the earlier "no cache exists yet" explanation was wrong, and the one I offered as the alternative is
the right one: **the restore path never finds anything.** The mechanism is cache scoping. `actions/cache`
scopes a saved cache to the ref that saved it, with the default branch's caches shared downward. Every
ref that runs this build is one-shot: merge-queue refs are named per group
(`gh-readonly-queue/main/pr-N-<sha>`), and `main`'s own C++ job has been cancelled 25 times out of 25, so
the default branch never saves anything shareable. Each save is therefore visible only to a ref that will
never run again — which is also why the same-run rerun I measured at the start hit 100%: that was one ref
reading its own cache, the one case scoping allows.

That makes the change unable to help the builds that matter, and it is not a tuning problem: a bigger cap
or dropping the HIP launcher changes nothing, because the entries are never reachable. So I am reverting
it rather than keeping a step that adds time and asserts a saving.

Two things that would make ccache work here, for whoever wants to try, neither of which is this PR:
make `main`'s CI able to complete (it is cancelled by concurrency on merge cadence, so nothing ever saves
a default-branch cache), or add a job that populates the cache on `main` and does not depend on that run
finishing. Until one of those exists, the honest state of this workflow is no ccache.

Issue #2552, opened 2026-09-18
ops(ci): PR Agent's gitnexus index runs on the NPU box, and five workflows can only use that box


bong-water-water-bong:
## Summary

`pr-agent.yml` runs on `[self-hosted, pr-agent]`, a label carried by **both** registered runners — including
`strixhalo-pr-agent`, the only box with an NPU. Five workflows require that box specifically
(`end-to-end-smoke.yml`, `kernel-object-lint.yml`, `npu-reset.yml`, `bundle-decoder-check.yml`, and
`bench.yml` after #2538 pins it), so a review bot's code-indexing step is competing with the only
hardware-dependent work in the repo.

## Measured

Both runners report `busy=true` while the NPU work queues. Identifying what holds `strixhalo` — the API
did not name it, so I looked at the box, which is local:

```
$ ps -eo pid,etime,cmd | grep Runner.Worker
1019639  00:16  .../Runner.Worker spawnclient 158 161
1019931  00:14  /usr/bin/bash -e .../_work/_temp/9fe0c444-....sh

$ grep jobDisplayName ~/actions-runner/_diag/Worker_20260918-031123-utc.log
  "jobDisplayName": "pr_agent"

$ tail -12 .../_work/_temp/9fe0c444-....sh
  npm install -g gitnexus@1.6.10-rc.204 2>/dev/null || true
  gitnexus index . 2>/dev/null
  gitnexus analyze || gitnexus analyze
  gitnexus detect-changes --scope compare --base-ref "origin/main" --repo 1bit-MONSTER
```

So it is the **gitnexus index/analyze step** of `pr_agent`, and it had been running for 16 minutes at the
time. Behind it, queued on that box: an `NPU Benchmark` run (mine, `#2538`'s verification, queued since
02:56) and a `Bundle decoder check` (queued since 02:56).

Consumption over the previous ~6h, from the run list, for scale: PR Agent **30 runs / 131 wall-min**
against NPU Benchmark 18 / 29 and End-to-End Smoke 11 / 57. Both figures are sums of run durations, not
box-seconds, so they overstate concurrency — but the shape is right: the bot is the second-largest
self-hosted consumer, and it has no hardware requirement.

## Why the NPU box is the one that suffers

`strixhalo-pr-agent` is the only host with `/opt/xilinx/xrt` and the NPU. `ryzen-pr-agent` has no XRT
[...]

PR #2555, opened 2026-09-18
ci(bench): the "Update BENCHMARKS.md timestamp" step could never run

bong-water-water-bong:
## What

`bench.yml` ended with a step that stamped a "Benchmarks run <now> — CI-verified"
line into `engine/npu/BENCHMARKS.md` via `sed -i`, and reported success. It could
not change the repository — or even the workspace copy — for three independent
reasons. Introduced in 200e4694a (2026-08-21, #1757), never functional.

**1. Nothing could land it.** `bench.yml` declares `permissions: contents: read`
(line 21) and there is no commit or push step:

    $ git log -S 'git commit' --format='%h %s' origin/main -- .github/workflows/bench.yml
    (empty)

The 8 workflows in this repo that *do* write back to it (census-autopr,
census-sweep, cf-analytics-bootstrap, cf-web-analytics, post-seo, seo-sync,
significant-post, update-badges) all declare a write permission. This one never
did, so even a successful edit is discarded with the runner workspace.

**2. The pattern matches nothing.** The `sed` pattern is `^\*Benchmarks`, and no
line in the tracked file starts with that:

    $ git log -S '*Benchmarks run' --format='%h %s' origin/main -- engine/npu/BENCHMARKS.md
    (empty)
    $ git show origin/main:engine/npu/BENCHMARKS.md | grep -cE '^\*Benchmarks'
    0

So `sed -i` exited 0 having changed nothing — a no-op that is indistinguishable
from a successful stamp in the run log (`sed -i` does not fail when the pattern
is absent, only when the file is unreadable).

**3. It was gated on an env flag.** `if: env.DRIVER_OK == 'true'` skipped it
whenever the driver was missing — a step recorded as a skip and read as a pass,
which is the failure shape #2538 fixes elsewhere in this same file.

Nothing else in the repo depends on the stamp: the only occurrence of
`Benchmarks run` anywhere is the `sed` itself, no workflow downloads the
`benchmark-log` artifact, and `Testing/npu_build_script_selfcheck.sh` (which does
parse `bench.yml`) is unaffected.

## The change

The last step becomes:

- gated on `always()` instead of `env.DRIVER_OK`, so an earlier failure is still
  recorded rather than silently skipped;
- writing to `$GITHUB_STEP_SUMMARY`, which is a record CI can actually write
  without a write token or a commit step;
- stating which of the three outcomes happened, rather than only the good one:
  `measured: <line>`, `no ms/tok line in benchmark.log — this run produced no
  measurement`, or `no benchmark.log — the engine never ran`.

I did not try to keep the tracked-file stamp by adding a write token plus a
commit step: a PR-triggered workflow pushing to `main` is a loop this repo
deliberately avoids, and `BENCHMARKS.md`'s own header already declares it
historical ("Status as of 2026-07-20 ... most of it dates to July 3").

## Verification

- The workflow still parses as YAML: 7 steps in the `benchmark` job, the old
  step name gone, exactly one new step.
- All three summary paths were exercised by running the step's own `run` text
  (`yaml.safe_load` -> `bash -c`, temp `GITHUB_STEP_SUMMARY`) with a log that
  contains `ms/tok`, a log without it, and no log at all. Each exits 0 and writes
  a summary naming the right outcome.
- Negative control: the check "the new step exists exactly once" was shown to
  fail when the step is renamed and when it is removed.
- `Testing/npu_build_script_selfcheck.sh` passes on the modified tree (rc=0),
  including its own controls.
- `bench.yml`'s push filter includes its own path, so this PR triggers a real
  benchmark run on `strix-halo`; that run exercises the new step.

Not claimed: that a previous run ever rendered this stamp (it could not have),
and that the summary is the *right* long-term home for benchmark provenance — if
the numbers should be machine-read into `site/benchmarks.json`, that is a
separate decision and a separate PR.

___

### **PR Type**
Bug fix

___

### **Description**
- Remove no-op `sed -i` BENCHMARKS.md stamp step

- Record verification timestamp in `$GITHUB_STEP_SUMMARY` instead

- Run step with `if: always()` rather than `env.DRIVER_OK`

- Report measured, no `ms/tok` line, or missing log

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["old step: Update BENCHMARKS.md timestamp"] -- "no write token, pattern never matched, skipped on DRIVER_OK" --> B["removed"]
  C["new step: Record the verification timestamp"] -- "if: always()" --> D["GITHUB_STEP_SUMMARY: measured / no ms-tok / no log"]
```

<tr>
  <td>
    <details>
      <summary><strong>bench.yml</strong><dd><code>Replace phantom BENCHMARKS.md stamp with job summary</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/bench.yml

<ul><li>Deletes the <code>Update BENCHMARKS.md timestamp</code> step, which <code>sed -i</code>'d a <br>CI-verified line into a tracked file with <code>permissions: contents: read</code> <br>and no commit step, and whose <code>^\*Benchmarks</code> pattern matched nothing.<br> <li> Adds a <code>Record the verification timestamp</code> step gated on <code>always()</code> <br>instead of <code>env.DRIVER_OK == 'true'</code>, so an earlier failure still <br>produces a record.<br> <li> Writes a timestamped heading plus one of three outcomes to <br><code>$GITHUB_STEP_SUMMARY</code>: the measured <code>ms/tok</code> line, "no ms/tok line in <br>benchmark.log", or "no benchmark.log — the engine never ran".<br> <li> Adds a comment block documenting the three independent reasons the old <br>step could never take effect.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2555/files#diff-a5e740be96415373789689f814583e93ff2a8f05eae6481e94505fd6cb6bc6a7">+31/-4</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Scanned all 32 workflow files to check whether this is one instance of a pattern
or the only one. Parsed each file with `yaml.safe_load`, walked every step's
`run`, flagged (a) `sed -i` against a path tracked in `origin/main` and (b)
`git commit`/`git push`, alongside whether the workflow or job declares a write
permission:

    bench.yml         job=benchmark  step=Update BENCHMARKS.md timestamp   sed -i           engine/npu/BENCHMARKS.md  write-perm=NO
    census-autopr.yml job=autopr     step=Open draft PR ...                git commit/push  -                         write-perm=YES
    census-sweep.yml  job=sweep      step=Open PR when the census moved    git commit/push  -                         write-perm=YES
    cf-analytics-bootstrap.yml  ...                                        git commit/push  -                         write-perm=YES
    cf-web-analytics.yml  ...                                              git commit/push  -                         write-perm=YES
    post-seo.yml      job=seo-tags   step=Open PR when post tags changed   git commit/push  -                         write-perm=YES
    seo-sync.yml      job=seo-sync   step=Open PR when drift found         git commit/push  -                         write-perm=YES
    significant-post.yml job=post    step=Open draft PR                    git commit/push  -                         write-perm=YES
    update-badges.yml job=update     step=Open a PR with badge updates     git commit/push  -                         write-perm=YES
    ci.yml            job=cpp        step=Build llama.cpp ggml-vulkan libs sed -i           src/CMakeLists.txt        write-perm=NO

One false positive, checked rather than assumed: `ci.yml`'s `sed -i
'/llama-onnx\.cpp/d' src/CMakeLists.txt` is preceded by `cd third_party/llama.cpp`
(line 89), so it patches the vendored submodule copy in the runner workspace —
deliberately local, not a tracked path of this repo (`src/CMakeLists.txt` does not
exist at this repo's root). Not a defect.

So the pattern is unique to the step this PR removes: every other workflow that
writes to the repo declares a write permission and commits. That convention is
also why I did not "fix" this by adding a write token to `bench.yml`.

bong-water-water-bong:
Evidence boundary for the last bullet, so the PR body is not read as more than it
is: the triggered run is **35303657605**, and at the time of writing it is still
`status=queued` — both self-hosted runners report `busy=true`, which is the
contention tracked in #2552. The push filter firing is a fact about the trigger;
the *rendered* summary is not yet observed, and I am not claiming it.

Verified from the repository itself rather than the working copy: the committed
blob of `bench.yml` parses as YAML with 7 steps in the `benchmark` job, the
phantom step name is absent, and `Record the verification timestamp` appears
exactly once. `runs-on: self-hosted` and `permissions: contents: read` are
untouched by this PR (the first is #2538's change, the second is load-bearing for
the argument here).

Also worth noting for review: `Testing/run_all.sh:356` already runs
`Testing/npu_build_script_selfcheck.sh`, which parses this same file, and that job
is green on this PR — so the edit is covered by an existing wired check, not only
by the ad-hoc test described above.

Issue #2486, opened 2026-09-17
ci(lint): a REQUIRED check cannot fail — 'Lint (clang-format)' is continue-on-error, so it enforces nothing while the ruleset lists it as required


bong-water-water-bong:
## Summary

`main`'s ruleset requires four status checks. Three of them do real work. **`Lint (clang-format)` cannot fail**, so it enforces nothing while appearing in the required list as if it did.

```
ruleset "main" (id 19117606, enforcement=active)
  required: C++ (cmake configure + build), Lint (clang-format), Scope Guard, Version consistency
```

`.github/workflows/ci.yml:205-222`:

```yaml
  lint:
    name: Lint (clang-format)          # <- the required context string
    runs-on: ubuntu-24.04
    steps:
      - name: clang-format check (informational)
        continue-on-error: true        # <- step-level: a failure does not fail the job
        run: |
          ...
          xargs clang-format --dry-run --Werror 2>&1 | head -20
```

Either of those two mechanisms alone is enough: `continue-on-error` keeps a step failure from failing the job, and the `| head -20` discards `xargs`' status regardless. So the job's conclusion is always success, and a *required* check that always succeeds gates nothing.

## Why this is filed as a contradiction rather than a bug in one place

The workflow's intent is explicit and reasoned — `ci.yml:211-215`:

> The `| head -20` below discards the xargs status, so this step could never fail regardless of its label — a latent trap for whoever promotes it by deleting "(informational)" … `continue-on-error` makes the intent EXPLICIT so removing the label is no longer what keeps it non-blocking.

So somebody deliberately made it *structurally* informational, precisely so that a later label change could not accidentally make it blocking. That is a good decision. But the ruleset then lists it as **required**, which says the opposite: a required check is one whose failure blocks a merge, and this one has no failing state. One of the two configurations is wrong, and they are in different files so neither author would have seen the other.

Consequence today: a PR that adds badly formatted C++ shows a green `Lint (clang-format)` required check. Formatting is not gated, and the check's presence in the required list actively suggests it is.

## The other three required checks do work — checked, not assumed

| required check | verdict |
|---|---|
| `C++ (cmake configure + build)` | real — it builds, and it is the slow one (5–17 min on recent PRs) |
| `Scope Guard` | real — I replayed its classification with its own extracted `ALLOWED` list: `docs`/`src`/`tools`/`engine`/`.github`/`hw` pass, `evilthing`/`unknownapp` are flagged as violations |
[...]

PR #2558, opened 2026-09-18
test(ci): all three red self-checks on main were reporting non-regressions

bong-water-water-bong:
### **User description**
`Self-checks (Testing/run_all.sh)` reports **three** failures on `main` right now, and
not one of them is a regression. This branch makes all three pass again, so the suite
can carry a signal at all.

Run directly on `origin/main` (`552ebae9b`):

| check | main | this branch |
|---|---|---|
| `npu_insts_lookup_selfcheck.sh` | FAIL (2) | PASS |
| `npu_build_script_selfcheck.sh` | FAIL | PASS |
| `capture_gates_selfcheck.py` | FAIL (2) | PASS |

Each failure, and what it actually was:

**1. `npu_insts_lookup` — a pattern pinned to a signature that changed.** The rule
located `ip()`'s lambda by the exact text `auto ip=[&](const char*t, int K, int N)`.
#2435 gave it default arguments and a trailing return type — `int K=-1, int N=-1) ->
std::string {` — so the awk range came back empty and the check reported "ip() has no
dimension-keyed fallback (regressed?)" about a file whose fallback was still there.
Its second failure was the same shape: `grep -nE '\bip\([^,()]*\)'` matched
`xp()/ip() prefer the` **inside a comment** at line 2133 — prose counted as a call
site. Used to pass at `75b28a8c1` and at the commit that introduced it (`02b4e04a3`);
red from `091699210` onwards.

**2. `npu_build_script` — a comment counted as a source.** `n_target` is counted
without deduping:

    awk '/set\(NPU_ENGINE_SOURCES/,/^\)/' CMakeLists.txt | grep -oE '[A-Za-z0-9_]+\.cpp' | wc -l

and that block says "model.c is C; runtime_layer.cpp + bridge are C++" — a source
named in prose. So `n_target` was 7 against 6 real sources, `n_ref` came out 1, and
the verdict skipped its "names no source, no copy to drift" branch for the "hand list"
FAIL — about a workflow that delegates to the CMake target and names nothing.

**3. `capture_gates` — one real drift and one control that could not fail.**
The marker was genuinely stale (`capture-gates-read` listed 2 names, the file
`getenv()`s 9). And the control "an emptied refused[] list is caught" asserted a catch
that is impossible by construction: every name in `refused[]` is also read via
`getenv()`, so coverage alone never notices — the control was reporting the truth
about its own premise. It now constructs an emptied list and asserts the rule that
*does* catch the edit fires.

## Controls

Every rule fixed here is one that had already gone blind, so each gets a control that
must fail when its subject is mutated:

- `ip()`/`xp()`: delete the fallback line → must be caught; rename the lambda → must be
  caught; append a comment naming `ip()` → must **not** be read as a call site.
- `NPU_ENGINE_SOURCES`: a commented name (and a duplicate of a real one) must not be
  counted; a real added entry must be.
- `capture_gates`: the emptied-list control now exercises the marker rule; verified it
  still catches a stale marker, and that a source whose `refused[]` is already empty
  reports exactly its two substantive failures without being misattributed to a
  control.

Also verified end-to-end against the real source, not only the copies: rewriting `ip()`
back to the pre-#2435 signature passes (the pattern is no longer signature-brittle),
and that signature *with* the fallback deleted is still caught.

## Scope note

`Self-checks` is not one of the four required checks, which is why a permanently red
suite blocked nothing and went unnoticed — the same gap as #2550 from the other side.
Making it required is a separate decision; this PR is only about the three rules that
could not report.

___

### **PR Type**
Tests, Bug fix

___

### **Description**
- Repair three self-checks that were permanently red on `main`

- `npu_insts_lookup`: find `ip()`/`xp()` by name, strip comments

- `npu_build_script`: strip comments and dedupe source lists

- `capture_gates`: correct getenv marker and its dead control

- Add mutation controls so every repaired rule can still fail

___

### Diagram Walkthrough

```mermaid
flowchart LR
  IS["npu_insts_lookup_selfcheck.sh"] -- "lambda_body name anchor + comment strip" --> SRC["ip()/xp() source rules"]
  BS["npu_build_script_selfcheck.sh"] -- "strip comments, sort -u" --> SRC2["NPU_ENGINE_SOURCES verdict"]
  CG["capture_gates_selfcheck.py"] -- "constructed empty refused[]" --> CTL["dead control now fires"]
  CI["cap_interposer.cpp"] -- "9-name getenv marker" --> CG
  IS -- "3 mutation controls" --> CTL
  BS -- "prose / added-entry controls" --> CTL
```

<tr>
  <td>
    <details>
      <summary><strong>npu_insts_lookup_selfcheck.sh</strong><dd><code>Anchor ip()/xp() self-checks by name, not signature text</code>&nbsp; </dd></summary>
<hr>

Testing/npu_insts_lookup_selfcheck.sh

<ul><li>Replace the exact-signature awk pins with <code>lambda_body</code>/<code>have_fallback</code> <br>helpers that locate the lambda by name and ignore comment lines.<br> <li> Add <code>onearg_sites</code>, a Python comment/string stripper that only reports <br>real <code>ip(...)</code> call sites instead of prose at line 2133.<br> <li> Add controls (a) removed fallback line, (b) renamed lambda, (c) <br>comment naming <code>ip()</code> — each must fail the rule.<br> <li> Guard the <code>sed</code> mutation with a <code>grep -cF</code> check so a stale pattern is <br>reported rather than silently passing.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2558/files#diff-1a611a3fdeea71f556d8d431b8f4f52aa60c4e3025970198f09c8cc13a273aec">+93/-6</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>npu_build_script_selfcheck.sh</strong><dd><code>Count NPU_ENGINE_SOURCES entries, not their prose</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/npu_build_script_selfcheck.sh

<ul><li>Add <code>npu_engine_sources()</code> which strips line comments and <code>sort </code><br><code>-u</code>-dedupes the <code>.cpp</code> names in <code>NPU_ENGINE_SOURCES</code>.<br> <li> Reuse it from <code>engine_sources_missing</code> and <code>sources_verdict</code> so prose like <br>"model.c is C; runtime_layer.cpp" is no longer counted.<br> <li> Strip comments and dedupe <code>ENGINE_OBJS</code> parsing in <code>objs_verdict</code>.<br> <li> Add control (d): a commented source name is invisible while a real <br>added entry is still counted (4).</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2558/files#diff-f7fb417ecf91b661b4d8877aca50a86e8f90c6a03ff16956de0514a4f397e1aa">+49/-5</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>capture_gates_selfcheck.py</strong><dd><code>Make the emptied-refused-list control able to fail</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Testing/capture_gates_selfcheck.py

<ul><li>Update the docstring example marker to the current nine-name <br><code>capture-gates-read</code> line and note the drift it caused.<br> <li> Rewrite the "emptied <code>refused[]</code> list" control to build the emptied list <br>from the source and assert the marker rule fires, since coverage alone <br>cannot detect it.<br> <li> Keep the shipped-text path from producing two misattributed control <br>failures when the source is already empty.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2558/files#diff-f2f36836caaf8e6106e7d2b94673aa754c4dc61346184690d35ccab9e5cf2ef9">+20/-3</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Documentation</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>cap_interposer.cpp</strong><dd><code>Sync capture-gates marker with the getenv set</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

npu-infer/tools/capture/cap_interposer.cpp

<ul><li>Expand the <code>capture-gates-read</code> marker to all nine <code>CAP_*</code> names the file <br><code>getenv()</code>s (<code>CAP_BIG_MAX</code>, <code>CAP_MM_W</code>, <code>CAP_NO_SYNC</code>, <code>CAP_RUNLIST_KV</code>, <br><code>CAP_POSTRUN_KV</code>, <code>CAP_DUMP_BIG</code>, <code>CAP_SKIP_BIG</code>, plus the documented <br><code>CAP_DIR</code>, <code>CAP_POSTRUN_ACT</code>).<br> <li> Update the header comment to distinguish documented operator gates <br>from value-style knobs and to require marker updates in the same <br>commit.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2558/files#diff-810bac0116ad0e0ee15ea0b3efeb8f5b1f39d04fd9c1429dec95b1a3f8c1dd84">+7/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
## Measured impact while this sat in the queue: every merge-group build has been red, and only on this

Checking the queue's health while these four entries waited, I found that **all eleven**
completed merge-group `CI` runs in the preceding hour concluded `failure`, and in each one the
failing job is **this PR's subject**:

    03:16:27  failure  35302584575   pr-2435
    03:18:01  failure  35302681458   pr-2507
    03:21:37  failure  35302907040   pr-2536
    03:22:11  failure  35302943337   pr-2550
    03:27:11  failure  35303257577   pr-2536
    03:30:16  failure  35303445772   pr-2551
    03:38:43  failure  35303975667   pr-2538
    03:44:52  failure  35304358320   pr-2553
    03:48:23  failure  35304579465   pr-2554
    03:54:11  failure  35304933525   pr-2555
    04:01:45  failure  35305392151   pr-2557

    $ for r in 35305392151 35304933525 35304579465; do
        gh api …/runs/$r/jobs | jq '.jobs[] | select(.conclusion=="failure") | .name'
      done
    Self-checks (Testing/run_all.sh)      # each run: exactly this job, nothing else

Two things worth separating:

* **It never blocked a merge**, because `Self-checks` is not one of the four required checks —
  `main` advanced through that whole window (#2538, #2553, #2554, #2555, #2557 all landed).
  So the cost was signal, not throughput: eleven consecutive red integration builds, each
  looking like a failure of whatever content it carried.
* **It was not noise anyone could dismiss from the outside.** The failing assertions were
  `npu_build_script`, `npu_insts_lookup` and `capture_gates_selfcheck` — the three this PR
  repairs — so every one of those runs was reporting a defect in the *changes it carried* that
  belonged to `main` instead.

**The falsifiable version of the fix:** the next merge-group `CI` runs after this lands should
conclude `success`. That is checkable from the run list without trusting my local runs — and if
a batch build is still red afterwards, this PR is not the whole story and I will say so.

bong-water-water-bong:
Strengthened from a 3-run sample to all of them: across the **11** red merge-group `CI` runs listed above, the failing job set is `Self-checks (Testing/run_all.sh)` in **11 of 11**, and in each run it is the *only* failure. So the prediction above is not a guess — there is no second red for this PR to miss.

The general shape, worth one line because it is the mirror of #2486: a check that is **not required** can be red on every integration build for hours without blocking anything *and* without anyone being told.  failing was indistinguishable, from the run list, from a real failure in the content each batch carried. #2486 is a required check that cannot fail; this was a non-required one that could not pass. This PR fixes the instance; whether the repository wants a guard that notices a check reddening  for K consecutive builds is a separate call, and one I would rather the owner make than add as another always-on alert.

bong-water-water-bong:
Correction to the comment above: its second paragraph lost two words, because I passed the body
inline in a shell double-quoted string with unescaped backticks, so bash executed `Self-checks`
and `main` as commands and substituted empty output. The repository's own `AGENTS.md` documents
exactly this failure ("deletes the text silently and leaves ZERO backticks"), and the answer is
the one written there: pass the body with `-F`. Here is the paragraph as intended:

> The general shape, worth one line because it is the mirror of #2486: a check that is **not
> required** can be red on every integration build for hours without blocking anything *and*
> without anyone being told. `Self-checks` failing was indistinguishable, from the run list, from
> a real failure in the content each batch carried. #2486 is a required check that cannot fail;
> this was a non-required one that could not pass. This PR fixes the instance; whether the
> repository wants a guard that notices a check reddening `main` for K consecutive builds is a
> separate call, and one I would rather the owner make than add as another always-on alert.

I also swept my other comments from the last three hours for the same damage: the rest are intact,
because everywhere else I escaped the backticks or wrote the body to a file first. This one was
the exception, and it is worth recording precisely because the failure is invisible in the posted
text — the sentence reads fine, it just means something else now.

bong-water-water-bong:
## Landed, and the prediction held

`main` is now `fb4e6e9ae` (*test(ci): all three red self-checks on main were reporting
non-regressions*), and both halves of the check I posted above came out as stated:

* **The merge-group CI run for the batch concluded `success`** — run `35306675265`, the group that
  carried these entries. That is the falsifiable claim from my earlier comment ("the next
  merge-group `CI` runs after this lands should conclude `success`"), after eleven consecutive
  `failure` groups in the preceding hour, each failing solely on this job.
* **The three checks pass on the new `main` when run directly**, not only inside the batch:

      npu_insts_lookup_selfcheck.sh    PASS
      npu_build_script_selfcheck.sh    PASS
      capture_gates_selfcheck.py       PASS

    (run from a clean worktree at `origin/main`, the same scripts CI runs.)

So a suite that had been red on every integration build for hours now reports, and the fix is
verified where it matters rather than only in the PR's own CI.

Still open from the parent issue, unchanged by this: #2486's decision (a *required* check that
cannot fail) and #2505's structural question (one build described in three places). This PR
fixed the third thing — three checks that could not pass — and nothing else.

Issue #2556, opened 2026-09-18
ci(bench): the benchmark TDR precondition (amdxdna timeout_in_sec=15) lives only in an untracked /etc file, and npu-reset.yml can silently revert it


bong-water-water-bong:
## Summary

Every 35B/dense measurement this repo has taken since 2026-09-16 carries the
precondition `amdxdna timeout_in_sec=15` (raised from the driver's 2 s default).
That precondition exists in exactly two places: prose in `benchmarks/*.md`, and an
untracked file in `/etc` on this one box. **Nothing in the repo sets it, and nothing
in CI reads it** — including the workflow that produces the CI benchmark numbers.

Two consequences:

1. On a box without that `/etc` file, the same benchmark workflow silently measures
   under the 2 s default — the condition the repo's own root-cause doc attributes the
   83-dump `ERT_CMD_STATE_TIMEOUT` teardown to.
2. `npu-reset.yml` reloads the module (lines 24-25: `rmmod amdxdna` / `modprobe
   amdxdna`), so it re-reads `modprobe.d`. Where the file is absent, that reset
   reverts the repair without saying so — and the next benchmark run is a different
   environment than the last.

## What is actually on the box (checked today)

```
$ cat /sys/module/amdxdna/parameters/timeout_in_sec
15
$ ls -la /etc/modprobe.d/ | grep -iE 'amdxdna|xdna'
-rw-r--r-- 1 root root  34 Sep 16 00:19 amdxdna-tdr.conf
-rw-r--r-- 1 root root 252 Sep  5 18:50 amdxdna.conf
$ modprobe -c | grep -i amdxdna
options amdxdna timeout_in_sec=15
options amdxdna context_limit=64
alias pci:v00001022d*sv*sd*bc11sc80i* amdxdna
```

The module was loaded `2026-09-17 12:12`, *after* `amdxdna-tdr.conf` was written
(`Sep 16 00:19`), and the live parameter reads 15 — so the reload path really does
pick the value up from `modprobe.d`. That file is the whole repair.

## Nothing in the repo reads or writes it

```
$ git grep -l 'timeout_in_sec' origin/main -- '*.sh' '*.yml' '*.py'
(empty)
$ git grep -c 'timeout_in_sec' origin/main -- 'benchmarks/*.md'
benchmarks/LEVERS-register-2026-09-15.md:1
benchmarks/RESULTS-consolidation-2026-09-16.md:1
[...]

PR #2560, opened 2026-09-18
ci(bench): assert the benchmark's TDR precondition instead of assuming it

bong-water-water-bong:
### **User description**
Addresses #2556 (the assertion half; the recording half is #2557).

ci(bench): assert the benchmark's TDR precondition instead of assuming it

The assertion half of #2556. Every 35B/dense number published since 2026-09-16 is
annotated timeout_in_sec=15; the driver's 2 s default is the condition
benchmarks/RESULTS-ert-rootcause-and-repair-2026-09-16.md attributes the 83-dump
ERT_CMD_STATE_TIMEOUT teardown to. That value lives in an untracked
/etc/modprobe.d/amdxdna-tdr.conf on one box, and nothing in the repo establishes it:

  $ git grep -l 'timeout_in_sec' origin/main -- '*.sh' '*.yml' '*.py'
  (empty)

so a runner without the file measures a different environment than the numbers were
taken in, and says nothing about it. bench.yml's driver pre-flight checked that the
driver is PRESENT, not that the environment is the one the published numbers describe.

The new step, between `Check NPU driver` and `Check XRT`:

  * reads /sys/module/amdxdna/parameters/timeout_in_sec (path overridable through
    TDR_PARAM, so the failure path is testable without touching the host);
  * puts the value and the kernel in the job summary, so the benchmark artifact is
    self-describing;
  * prints the override's own state, because an override nobody notices is how a
    precondition quietly becomes advisory;
  * fails closed when the value is the 2 s default or unreadable, unless the new
    allow_default_tdr dispatch input is set — and then it warns and says its numbers
    are not like-for-like.

Verified by running the step's own `run` text against fixtures rather than the host
parameter:

    fixture=15, no override        rc=0   no error/warning   "precondition met"
    fixture=2,  no override        rc=1   ::error:: naming this issue
    fixture=2,  allow_default_tdr  rc=0   ::warning:: numbers not comparable
    unreadable path                rc=1   ::error::   (value reads '?')

and in each case the summary table is written. Step order asserted, not assumed:
index 2, after Check NPU driver and before Check XRT.

Not verified here, stated so it is not read as measured: the Actions expression
wiring for non-dispatch events (inputs.allow_default_tdr == true && '1' || '' is
documented to evaluate to '' on push). The step prints the effective value on every
run, so the first push-triggered run shows it — and bench.yml's own path is in its
push filter, so this PR triggers exactly that run.

The recording half of #2556 is #2557 (npu-reset.yml). merge-tree against this file's
other open PR (#2555, which edits the tail) is CLEAN — the hunks are disjoint.

___

### **PR Type**
Enhancement

___

### **Description**
- Add fail-closed TDR precondition check to `bench.yml`

- Read `amdxdna` `timeout_in_sec` and print it

- Record value and kernel in job summary

- New `allow_default_tdr` dispatch input to bypass, with warning

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["workflow_dispatch input allow_default_tdr"] -- "boolean, default false" --> B["Check the measurement precondition"]
  B -- "cat TDR_PARAM / timeout_in_sec" --> C["tdr value: 15 / 2 / ?"]
  C -- "15" --> D["precondition met, step passes"]
  C -- "2 or unreadable" --> E["::error:: fail closed (#2556)"]
  C -- "2 + override set" --> F["::warning:: not like-for-like, continue"]
  C -- "always" --> G["GITHUB_STEP_SUMMARY: kernel + timeout_in_sec"]
  B -- "runs between" --> H["Check NPU driver"]
  B -- "and" --> I["Check XRT"]
```

<tr>
  <td>
    <details>
      <summary><strong>bench.yml</strong><dd><code>Assert amdxdna TDR timeout before benchmark run</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/bench.yml

<ul><li>Adds a <code>Check the measurement precondition (amdxdna TDR)</code> step between <br><code>Check NPU driver</code> and <code>Check XRT</code>, reading <code>TDR_PARAM</code> (default <br><code>/sys/module/amdxdna/parameters/timeout_in_sec</code>).<br> <li> Prints the read value, the override's own state, and appends a <br>markdown table with the kernel and <code>timeout_in_sec</code> to <br><code>$GITHUB_STEP_SUMMARY</code> for self-describing artifacts.<br> <li> Fails closed with <code>::error::</code> when the value is the 2 s default or <br>unreadable, referencing #2556; warns instead when overridden.<br> <li> Adds a new <code>allow_default_tdr</code> boolean <code>workflow_dispatch</code> input (default <br><code>false</code>) mapped to <code>ALLOW_DEFAULT_TDR</code>, documented as the deliberate-2 s <br>escape hatch.</ul>

</details>

  </td>
  <td><a href="https://github.com/1bit-MONSTER/1bit-MONSTER/pull/2560/files#diff-a5e740be96415373789689f814583e93ff2a8f05eae6481e94505fd6cb6bc6a7">+47/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


bong-water-water-bong:
Verification matrix completed — the production path, not only fixtures:

    TDR_PARAM unset  -> reads /sys/module/amdxdna/parameters/timeout_in_sec
    amdxdna timeout_in_sec = 15 (from /sys/module/amdxdna/parameters/timeout_in_sec)
    allow_default_tdr override: '<unset>'
    measurement precondition met (timeout_in_sec=15)   rc=0, summary written

So all five cases are exercised: `15` → pass, `2` → fail, `2` + override → pass with warning, unreadable → fail, and the **default path string** against the live parameter. The push-triggered CI run (35305299072) is still queued behind the self-hosted backlog, but the step itself no longer has an unexercised branch — only an unexercised *environment*, which is the runner's.

bong-water-water-bong:
## The step passed in CI on the real runner; the job's red is main's build break, not this change

Run 35305299072, job `benchmark`, runner `strixhalo-pr-agent`:

    amdxdna timeout_in_sec = 15 (from /sys/module/amdxdna/parameters/timeout_in_sec)
    allow_default_tdr override: '<unset>'
    measurement precondition met (timeout_in_sec=15)

and the failing step is **Build engine (continuous batch)** — 31 undefined references
(`flm_prefill_init`, …), i.e. the CMake engine target that **#2561** repairs and that has been
unable to link on `main` since #2435. #2561 sits in the next merge-queue batch, so this job
should go green end to end on the following run without any further change here.

The summary table this step writes is visible in the log's step script; the step itself exited 0
before the build ran, which is the ordering the PR relies on (assert the environment, then spend
N minutes building).

Issue #987, opened 2026-07-26
CI workflows: no third-party actions pinned to SHA — supply-chain attack vector


bong-water-water-bong:
## Summary

All 11 CI workflow files use third-party GitHub Actions via mutable version tags (e.g., `@v4`, `@v5`) instead of pinning to an immutable commit SHA. A tag can be force-pushed to point to malicious code without changing the workflow file.

## Affected workflows

| Workflow | Unpinned actions |
|----------|:---------------:|
| `.github/workflows/bench.yml` | 2 (`actions/checkout@v4`, `stefanzweifel/git-auto-commit-action@v5`) |
| `.github/workflows/ci.yml` | 5 (`actions/checkout@v4`, `actions/setup-python@v5` ×3, `codecov/codecov-action@v5`) |
| `.github/workflows/deploy.yml` | 2 (`actions/checkout@v4`, `peaceiris/actions-gh-pages@v4`) |
| `.github/workflows/pr-agent.yml` | 2 (`actions/checkout@v4`, `Codium-ai/pr-agent@main`) |
| `.github/workflows/release.yml` | 2 (`actions/checkout@v4`, `softprops/action-gh-release@v2`) |
| `.github/workflows/update-badges.yml` | 1 (`stefanzweifel/git-auto-commit-action@v5`) |
| `.github/workflows/validate-benchmarks.yml` | 1 (`actions/checkout@v4`) |
| `.github/workflows/validate-claims.yml` | 5 (`actions/checkout@v4`, `actions/setup-python@v5` ×3, `stefanzweifel/git-auto-commit-action@v5`) |
| `.github/workflows/video-lora-ci.yml` | 6 (3× `actions/checkout@v4`, 3× `actions/setup-python@v5`) |

## Risk

- `Codium-ai/pr-agent@main` (in `pr-agent.yml`) is especially risky — `@main` branch can change at any time. This action has `issues: write` and `pull-requests: write` permissions, meaning injected code could read/write issues and PRs.
- `peaceiris/actions-gh-pages@v4` (in `deploy.yml`) has `deployments: write` permission — malicious code could deploy arbitrary content to GitHub Pages.
- `stefanzweifel/git-auto-commit-action@v5` (in `bench.yml`, `validate-claims.yml`, `update-badges.yml`) has `contents: write` permission — injected code could push arbitrary commits to the repo.

## Suggested fix

Pin every third-party action to its current commit SHA and add a comment with the human-readable version:

[...]

PR #2565, opened 2026-09-18
ci(security): pin the one remaining action tag — ShellCheck runs in a required job

bong-water-water-bong:
Fixes the remaining instance of the policy in #987 — one `uses:` line in a required job.
ci(security): pin the one remaining action tag — ShellCheck runs in a required job

`ci.yml` used `ludeeus/action-shellcheck@2.0.0`, a mutable third-party tag, while every
other action in the repository is pinned to a commit. #987 (closed 2026-07-26) is the
audit that removed the mutable tags, and this usage arrived **after** it — added by
#1756 on 2026-08-21 — so it is drift from a policy that was already in place rather
than a case the audit missed.

It is worth more than bookkeeping because of where it runs: the ShellCheck step lives
in `C++ (cmake configure + build)`, one of the four checks the merge-queue ruleset
requires. A force-pushed or rewritten tag would execute arbitrary code under that gate,
with the job's token and the repository checkout. The exposure is not theoretical — it
is the entire point of #987.

The pin is 00cae500b08a931fb5698e11e79bfbd38e612a38, which is what `2.0.0` resolved to
when this step last ran: the job log's own download line prints that SHA
(`Download action repository 'ludeeus/action-shellcheck@2.0.0' (SHA:00cae500…)`), so the
commit pinned here is the code that has been executing, not a newer revision.

Verified: the workflow still parses with its 7 jobs, and a sweep of every `uses:` line
across the 34 workflow files now reports 56 pinned to a 40-hex commit and **0 on a
mutable ref** (it was 1 of 56).


bong-water-water-bong:
The proof this PR owes: the pinned SHA is the one that actually executes. It is, and the inline `# 2.0.0` annotation is accurate.

- Job `C++ (cmake configure + build)` = `105483915952` in run `35307951947` (18m18s) — step **`ShellCheck` `completed/success`**, and every action sub-step reports `outcome=success`.
- The step's `Run the check` invoked the binary from the pinned path: `/home/runner/work/_actions/ludeeus/action-shellcheck/00cae500b08a931fb5698e11e79bfbd38e612a38/shellcheck`, with `INPUT_SHELLCHECK_OPTIONS: -S warning --format=gcc`, taking 11.5s.
- The tool that ran is `ShellCheck - shell script analysis tool / version: 0.11.0`.
- Mapping confirmed independently: `git/ref/tags/2.0.0` resolves to commit `00cae500b08a931fb5698e11e79bfbd38e612a38`, and `releases/latest` is `2.0.0` (published 2023-01-29). The tag is still what `main` carries (`uses: ludeeus/action-shellcheck@2.0.0` at ci.yml:197); this PR is the change that replaces it.

All 16 checks on this head pass, `enqueue` included. A tag/SHA mismatch would fail at exactly this step — the action would not resolve — so the step outcome is the discriminator, not the presence of the `uses:` line.

PR #2629, opened 2026-09-21
fix(moe): correct the 35B expert dequant to the vendor layout (group-major + subtractive)

bong-water-water-bong:
The expert Q4NX dequant in both the engine's host-FFN path
(`runtime_layer_moe.cpp::dequant_1bp_engine`) and the numpy reference
(`tools/qwen36_full_ref.py::_exp_deq`) used ROW-major scales `sc[r*8+g]`,
ROW-major nibbles, and the ADDITIVE torch2aie form `q*scale+zp`.  Both agreed
with each other (corr 1.0) but are ORTHOGONAL to the vendor's own dequantiser.

Independent check against the library FLM itself links —
`Q4NX::q4nx_dequantize<biovault::bfloat16_t>` in `libq4_npu_eXpress.so`, called
through its exported symbol on real expert tiles from `model.q4nx` — settles the
layout:

- 256 bf16 scales + 256 bf16 zero-points, GROUP-major `sc[g*32 + row]`
  (g = col/32, row 0..31)
- int4 nibbles 16-lane interleaved: `byte = lane*2048 + col*8 + (row%16)/2`
- SUBTRACTIVE `W = (q - zp) * scale`

Vendor layout vs the vendor .so output: corr 0.999998 (max|diff| ~ bf16
rounding).  Previous row-major/additive layout: corr 0.011.

This is the self-referential defect the auditor flagged: the reference was
verified only against the engine's own (equally wrong) dequant.  `model.c`'s
`npu_dequant_block` already had the correct layout; only the MoE host path and
the numpy reference carried the wrong one.

Fixes:
- runtime_layer_moe.cpp::dequant_1bp_engine -> group-major + 16-lane + subtractive
- tools/qwen36_full_ref.py::_exp_deq -> group-major + 16-lane + subtractive
- model.c: correct the stale scale/formula header comment
- npu-infer/tests/verify_exp_deq_vs_vendor.cpp: the independent check (new)

Result (full 40-layer decode, token 151644): engine host-FFN path and the
corrected reference now agree bit-close (corr 0.999984, both greedy 234071),
previously they disagreed.  FLM (greedy 386) is still not matched (corr 0.039) —
the Q8_0 lm_head/attention/shared-expert dequant is the next suspect.

Ref: goal mu721yrt-p5wy9i (FastFlowLM 35B MoE parity), auditor objection #2.


bong-water-water-bong:
Superseded by #2630 (the same int4 Q4NX expert-dequant fix off current main), which is **merged** as 3a7651a4a. This branch diverged and conflicted; keeping it open only obscures the trail. The current campaign PRs are #2631 (Q8_0 int8 column-major), #2633 (ELF path vs corrected baseline + one retraction) and #2634 (host routed-FFN threading).

